PR #70, opened 2026-01-30
fix: Warn when config loading fails instead of silent fallback

jlaneve:
## Summary

- Print warning to stderr when config loading fails (e.g., missing env var)
- Update skill docs with missing setup commands and correct env var names

## Problem

When config file exists but has errors (like `${AIRFLOW_API_TOKEN}` when the env var isn't set), the CLI silently falls back to localhost:8080. Users see a connection error to localhost without understanding why their config wasn't used.

## Solution

Print a warning to stderr before falling back:
```
Warning: Failed to load config: Environment variable 'AIRFLOW_API_TOKEN' is not set
Falling back to default settings (localhost:8080)
```

## Skill Doc Fixes

- Added `af instance add` command (was missing entirely!)
- Fixed env var name: `AIRFLOW_CLI_CONFIG` → `AF_CONFIG`
- Documented `${VAR}` syntax for tokens in config
- Documented `--token` CLI flag
- Added example YAML config with env var interpolation

## Test Plan

- [x] Verified warning appears when env var is missing
- [x] Verified no warning when config file doesn't exist (first-time user)
- [x] Verified normal operation when config loads successfully

---

### DIFF
--- a/skills/airflow/SKILL.md
+++ b/skills/airflow/SKILL.md
@@ -22,6 +22,10 @@ Throughout this document, `af` is shorthand for `uvx --from astro-airflow-mcp@la
 Manage multiple Airflow instances with persistent configuration:
 
 ```bash
+# Add a new instance
+af instance add prod --url https://airflow.example.com --token "$API_TOKEN"
+af instance add staging --url https://staging.example.com --username admin --password admin
+
 # List and switch instances
 af instance list      # Shows all instances in a table
 af instance use prod  # Switch to prod instance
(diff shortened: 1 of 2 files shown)


skills/airflow/SKILL.md
```diff
@@ -22,6 +22,10 @@ Throughout this document, `af` is shorthand for `uvx --from astro-airflow-mcp@la
 Manage multiple Airflow instances with persistent configuration:
 
 ```bash
+# Add a new instance
+af instance add prod --url https://airflow.example.com --token "$API_TOKEN"
+af instance add staging --url https://staging.example.com --username admin --password admin
+
```
schnie, line 28: Did we want the agent to do adding? Yesterday we discussed not but, I have auto-discovery coming and it could use that too.

skills/airflow/SKILL.md
```diff
@@ -32,17 +36,28 @@ af instance delete old-instance
...
 ```bash
 export AIRFLOW_API_URL=http://localhost:8080
+export AIRFLOW_AUTH_TOKEN=your-token-here
+# Or username/password:
 export AIRFLOW_USERNAME=admin
 export AIRFLOW_PASSWORD=admin
 ```
 
-Or CLI flags: `af --airflow-url http://localhost:8080 --username admin --password admin <command>`
+Or CLI flags: `af --airflow-url http://localhost:8080 --token "$TOKEN" <command>`
```
schnie, line 60: I wonder if we should just drop this flag/env var usage from the skill in favor of `af instance` management.

skills/airflow/SKILL.md
jlaneve: I think it's fine to have the agent add stuff yeah, should be somewhat rare bc once it's set up we're good to go

skills/airflow/SKILL.md
jlaneve: hmm I was thinking about that, but for stuff like CI it's easier to set env vars vs write a local file

PR #98, opened 2026-02-04
Skill: Best practices

igorastronomer:
Updated skill: 
 - **authoring-dags/best-practices**: Added sections on avoiding top-level code, dynamic task mapping, XCom limits, retries/scaling, deferrable operators, and Airflow 3 Assets

### DIFF
--- a/skills/authoring-dags/reference/best-practices.md
+++ b/skills/authoring-dags/reference/best-practices.md
@@ -235,7 +235,7 @@ process_file.partial(output_dir="/out").expand(filename=get_files())
 
 ## Handle Large Data (XCom Limits)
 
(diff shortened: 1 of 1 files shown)


skills/authoring-dags/reference/best-practices.md
```diff
@@ -167,6 +206,109 @@ def extract_sources():
...
+
+# With constant parameters
+process_file.partial(output_dir="/out").expand(filename=get_files())
+```
+
+---
+
+## Handle Large Data (XCom Limits)
+
+XCom default limit is ~1MB. For large data, use external storage:
```
kaxil: Where is this 1MB limit coming from?

igorastronomer: Incorrectly assumed from the docs that the example indicated the limit. 

Updated it to: 

`Airflow does not define a single universal "max XCom size". XComs are designed for **small** values only (do not pass large payloads like dataframes). For large data, write to external/object storage and pass a reference (URI/path) via XCom. See: `https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/xcoms.html`.`

`For automatic offloading, use the Object Storage XCom backend (provider `common-io`): values larger than a configured threshold are stored in object storage, while smaller values remain in the metadata DB. The threshold is **configurable** (docs examples often use `1048576` bytes = 1 MiB)..`

PR #102, opened 2026-02-05
fix: loading hooks when running --plugin-dir

milton-li:
Running the plugin in Claude with `--plugin-dir` doesn't properly load in hooks for local development. This creates a plugin manifest with hook definition.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -17,19 +17,7 @@
       },
       "homepage": "https://github.com/astronomer/agents",
       "repository": "https://github.com/astronomer/agents",
-      "keywords": ["data-engineering", "airflow", "snowflake", "bigquery", "jupyter", "astronomer"],
-      "hooks": {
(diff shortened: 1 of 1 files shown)


.claude-plugin/plugin.json
```diff
@@ -0,0 +1,17 @@
+{
```
kaxil, line 1: This should have already worked with https://github.com/astronomer/agents/blob/main/.claude-plugin/marketplace.json though

milton-li: This works when installing from the marketplace distribution, but running the plugin locally with `--plugin-dir` doesn't load in the hooks defined in `marketplace.json`. Alternatively, we could create a `/hooks/hooks.json` which I've confirmed also works.

schnie: Can confirm. Need to do this with the SessionStart hook too.

schnie (approved):

PR #103, opened 2026-02-05
refactoring dags skill

RNHTTR:


skills/refactoring-dags/SKILL.md
```diff
@@ -0,0 +1,269 @@
...
+
+---
+
+## CRITICAL WARNING: Use MCP Tools, NOT CLI Commands
+
+> **STOP! Before running ANY Airflow-related command, read this.**
+>
+> You MUST use MCP tools for ALL Airflow interactions. CLI commands like `astro dev run`, `airflow dags`, or shell commands to read logs are **FORBIDDEN**.
+>
+> **Why?** MCP tools provide structured, reliable output. CLI commands are fragile, produce unstructured text, and often fail silently.
```
jlaneve, line 25: this is kind of counter to what we're trying to do with the `af` cli btw

RNHTTR: fwiw this is the exact same SKILL.md as the one for the existing create dags skill

PR #104, opened 2026-02-09
Improve authoring-dags skill based on early user feedback

tayloramurphy:
## Summary

Addresses feedback from an early adopter using the authoring-dags skill with Junie/PyCharm. The skill was producing some false positives and missing common patterns.

Changes to `skills/authoring-dags/reference/best-practices.md`:

- **Airflow 3 context injection**: Document that `data_interval_end: pendulum.DateTime` (no `= None`) is valid in AF3 Task SDK — the skill was flagging this as a missing argument
- **Claim-check pattern**: Name and describe the pattern explicitly so the skill recognizes it when already applied in existing code, not just when recommending it
- **XCom size context**: Add guidance to consider actual data volume from code context before flagging XCom concerns (e.g. a 12-row DataFrame is fine)
- **Sensor modes**: Expand from just `deferrable=True` to cover `reschedule` vs `poke` mode — this was actually a strength the user called out, now it's documented
- **Error handling/callbacks**: New section covering failure callbacks and trigger rules (gap vs the patterns users expect)
- **Outlets without inlets**: Clarify these are valid for lineage — the skill was adding outlets in one DAG then complaining about outlet-only assets in another

Also adds a matching note to `skills/annotating-task-lineage/SKILL.md` so the lineage skills stay consistent on the outlets-without-inlets point.

## Test plan

- [ ] Verify skill loads correctly with `claude --plugin-dir .`
- [ ] Run doctoc to confirm TOC is up to date (already ran)
- [ ] Review best-practices.md reads coherently end-to-end

### DIFF
--- a/skills/authoring-dags/reference/best-practices.md
+++ b/skills/authoring-dags/reference/best-practices.md
@@ -30,7 +30,7 @@ The examples below use Airflow 2 imports for compatibility. On Airflow 3, these 
 - [Dynamic Task Mapping](#use-dynamic-task-mapping)
 - [Large Data / XCom](#handle-large-data-xcom-limits)
 - [Retries and Scaling](#configure-retries-and-scaling)
-- [Deferrable Operators](#use-deferrable-operators)
+- [Sensor Modes and Deferrable Operators](#sensor-modes-and-deferrable-operators)
 - [Setup/Teardown](#use-setupteardown)
 - [Data Quality Checks](#include-data-quality-checks)
 - [Anti-Patterns](#anti-patterns)
@@ -162,6 +162,21 @@ sql = """
 """
 ```
 
+**Airflow 3 context injection**: In Airflow 3 (Task SDK), context variables are automatically injected as function parameters by name. A bare type annotation is valid — no `= None` default required:
+
+```python
+import pendulum
+
+# Airflow 3 — both forms are valid
+@task
+def process(data_interval_end: pendulum.DateTime):  # No default needed
+    ...
+
+@task
+def process(data_interval_end: pendulum.DateTime = None):  # Also valid but unnecessary in AF3
+    ...
+```
+
 ---
 
 ## Organize with Task Groups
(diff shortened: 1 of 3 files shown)


skills/authoring-dags/reference/best-practices.md
```diff
@@ -295,20 +325,106 @@ AIRFLOW__CORE__PARALLELISM=32
 
 ---
 
-## Use Deferrable Operators
+## Sensor Modes and Deferrable Operators
```
kaxil, line 291: I don't think we should put `reschedule` vs `poke` in best practises. Only Deferrable is fine imo.

cc @TJaniF

skills/authoring-dags/reference/best-practices.md
kaxil: Argh, doctoc shouldn't be in our skills! -- let me remove that in a separate PR/commit

kaxil: https://github.com/astronomer/agents/pull/107

skills/authoring-dags/reference/best-practices.md
```diff
@@ -237,25 +263,29 @@ process_file.partial(output_dir="/out").expand(filename=get_files())
...
+    return path  # Small string reference (the "claim check")
 
 @task
 def transform(path: str):
-    data = pd.read_parquet(path)
+    data = pd.read_parquet(path)  # Retrieve data using the reference
     ...
 ```
 
+**Context matters for XCom size concerns**: When reviewing code that returns data via XCom (e.g., a DataFrame), consider the actual data volume based on code context. A task that returns a DataFrame from an API with a known small result set (e.g., 12 rows of monthly aggregates) is not an XCom size concern. Only flag XCom size issues when the data volume is genuinely unbounded or likely to be large.
```
kaxil: This is oddly specific example :D I know why that is but I think we can change this section

skills/authoring-dags/reference/best-practices.md
```diff
@@ -295,20 +325,106 @@ AIRFLOW__CORE__PARALLELISM=32
...
+    message = f"Task {ti.task_id} failed in DAG {ti.dag_id} for {context['ds']}"
+    # Send to Slack, PagerDuty, email, etc.
+
+@dag(
+    schedule='@daily',
+    start_date=datetime(2025, 1, 1),
+    catchup=False,
+    on_failure_callback=task_failure_callback,  # DAG-level failure callback
+    default_args={
+        'on_failure_callback': task_failure_callback,  # Task-level default
```
kaxil: The example shows some callback used at Dag-level and task-level though! Why double

kaxil: There are also Notifiers: https://airflow.apache.org/docs/apache-airflow-providers/core-extensions/notifications.html

PR #106, opened 2026-02-09
Adds `af` CLI telemetry

schnie:
This PR instruments the `af` CLI with anonymous telemetry. Users can opt-out of this with `af telemetry disable`.

### DIFF
--- a/astro-airflow-mcp/README.md
+++ b/astro-airflow-mcp/README.md
@@ -444,7 +444,7 @@ echo astro-airflow-mcp >> requirements.txt
 
 ### Telemetry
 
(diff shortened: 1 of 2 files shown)


astro-airflow-mcp/README.md
```diff
@@ -442,6 +442,18 @@ echo astro-airflow-mcp >> requirements.txt
 | `--config`, `-c` | `AF_CONFIG` | Path to config file (default: `~/.af/config.yaml`) |
 | `--version`, `-v` | | Show version and exit |
 
+### Telemetry
+
+The `af` CLI collects anonymous usage telemetry to help improve the tool.
```
jlaneve: should we say more about what specifically it is?

PR #124, opened 2026-02-18
Add mandatory unpause step to testing-dags skill

josh-fell:
Adds a required unpause step before triggering DAGs to prevent them from getting stuck in queued state. This fixes a common issue where paused DAGs accept triggers but never execute due to Airflow's `dags_are_paused_at_creation` default configuration. Updates all examples and workflow diagrams to reflect this best practice and adds troubleshooting guidance for paused DAG scenarios.

### DIFF
--- a/skills/testing-dags/SKILL.md
+++ b/skills/testing-dags/SKILL.md
@@ -19,12 +19,34 @@ Throughout this document, `af` is shorthand for `uvx --from astro-airflow-mcp af
 
 ---
 
-## FIRST ACTION: Just Trigger the DAG
+## FIRST ACTION: Ensure DAG is Unpaused, Then Trigger
 
-When the user asks to test a DAG, your **FIRST AND ONLY action** should be:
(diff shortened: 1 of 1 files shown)


skills/testing-dags/SKILL.md
```diff
@@ -137,16 +171,19 @@ af runs trigger-wait my_dag --timeout 300
...
 
-### Alternative: Trigger and Monitor Separately
+### Alternative: Unpause + Trigger and Monitor Separately
 
 Use this only when you need more control:
 
 ```bash
-# Step 1: Trigger
+# Step 1: Unpause
+af dags unpause my_dag
```
kaxil, line 180: We could move that logic in `af runs trigger` command too --- so it is deterministic

josh-fell: Oh that tool is in the repo now. Agreed, much better idea. I'll close this PR and go with that approach instead.

PR #128, opened 2026-02-19
Removes write keys

schnie:
Removes the write keys from the code base.

### DIFF
--- a/astro-airflow-mcp/pyproject.toml
+++ b/astro-airflow-mcp/pyproject.toml
@@ -92,7 +92,6 @@ ignore = [
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401"]  # Allow unused imports in __init__.py
 "tests/**/*.py" = ["ARG001", "ARG002", "PT011"]  # Allow unused arguments and broad pytest.raises in tests
-"src/astro_airflow_mcp/cli/main.py" = ["E402"]  # Allow import ordering flexibility
 
 [tool.bandit]
 exclude_dirs = ["tests", "examples"]
--- a/astro-airflow-mcp/src/astro_airflow_mcp/cli/telemetry.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/cli/telemetry.py
@@ -193,22 +193,21 @@ d = json.loads(sys.stdin.read())
 data = json.dumps(d["body"]).encode("utf-8")
 req = request.Request(
     d["api_url"], data=data,
-    headers={{"Content-Type": "application/json"}}, method="POST",
+    headers={"Content-Type": "application/json"}, method="POST",
 )
 debug = d.get("debug", False)
 try:
-    with request.urlopen(req, timeout={timeout}) as resp:
+    with request.urlopen(req, timeout=__TIMEOUT__) as resp:
         body = resp.read().decode("utf-8", errors="replace")
         if debug:
-            print(f"[telemetry] response: {{resp.status}} {{body}}", file=sys.stderr)
+            print(f"[telemetry] response: {resp.status} {body}", file=sys.stderr)
 except error.HTTPError as e:
-    with e:
-        body = e.read().decode("utf-8", errors="replace")
-        if debug:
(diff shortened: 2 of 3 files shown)


astro-airflow-mcp/src/astro_airflow_mcp/cli/telemetry.py
```diff
@@ -200,31 +172,71 @@ def track_command() -> None:
...
+    if debug:
+        print(f"[telemetry] error: {{e}}", file=sys.stderr)
+"""
+
+
+def _send(api_url: str, body: dict, *, debug: bool = False) -> None:
+    """Send telemetry event in a detached subprocess.
+
+    When debug=True, logs request/response details to stderr and waits for
+    the subprocess to finish. Otherwise fire-and-forget.
```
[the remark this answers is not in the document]
schnie: Fixed

astro-airflow-mcp/pyproject.toml
```diff
@@ -93,7 +92,7 @@ ignore = [
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401"]  # Allow unused imports in __init__.py
 "tests/**/*.py" = ["ARG001", "ARG002", "PT011"]  # Allow unused arguments and broad pytest.raises in tests
-"src/astro_airflow_mcp/cli/main.py" = ["E402"]  # Warning filter must run before imports
+"src/astro_airflow_mcp/cli/main.py" = ["E402"]  # Allow import ordering flexibility
```
kaxil: Is this E402 ignore still needed? The `warnings.filterwarnings()` call that required it is gone now. If there's no remaining violation, let's just drop the ignore entirely.

astro-airflow-mcp/src/astro_airflow_mcp/cli/telemetry.py
```diff
@@ -200,31 +172,72 @@ def track_command() -> None:
...
+    headers={{"Content-Type": "application/json"}}, method="POST",
+)
+debug = d.get("debug", False)
+try:
+    with request.urlopen(req, timeout={timeout}) as resp:
+        body = resp.read().decode("utf-8", errors="replace")
+        if debug:
+            print(f"[telemetry] response: {{resp.status}} {{body}}", file=sys.stderr)
+except error.HTTPError as e:
+    with e:
```
kaxil: `with e:` on an exception object is pretty surprising — most people won't expect that to be valid. It works because `HTTPError` inherits from `HTTPResponse`, but it reads like a typo. Can we just do `e.close()` after the read, or skip closing entirely since this is a throwaway subprocess?

astro-airflow-mcp/src/astro_airflow_mcp/cli/telemetry.py
```diff
@@ -200,31 +172,72 @@ def track_command() -> None:
...
+        "source": TELEMETRY_SOURCE,
+        "event": "CLI Command",
+        "anonymousId": anonymous_id,
+        "properties": properties,
+    }
+
+    _send(api_url, body, debug=debug)
+
+
+_SEND_SCRIPT = """\
```
kaxil, line 188: Using `.format()` on a big code string means every literal brace has to be doubled — there are already 6+ `{{`/`}}` in here. If someone later adds a dict literal or f-string to this script and forgets to double the braces, it'll blow up at runtime with a `KeyError`, not at import time.

Worth switching to `string.Template("$timeout")` or a simple `.replace("__TIMEOUT__", ...)` to avoid that footgun.

schnie: Done!

astro-airflow-mcp/src/astro_airflow_mcp/cli/telemetry.py
schnie: Fixed.

astro-airflow-mcp/pyproject.toml
schnie: Nope, fixed. We removed that code.

PR #129, opened 2026-02-19
feat: add `af runs delete` and `af runs clear` commands

tayloramurphy:
## Summary

- Add first-class CLI commands and MCP tools for deleting and clearing DAG runs,
  replacing the need to use `af api` or one-off scripts for these common operations
- `af runs delete <dag_id> <run_id>` — permanently removes a DAG run (with confirmation prompt)
- `af runs clear <dag_id> <run_id>` — resets a DAG run for re-execution (with `--dry-run` support)
- Both commands are available as MCP tools (`delete_dag_run`, `clear_dag_run`) for AI assistant use

## Changes

- **Adapter layer**: Added `delete_dag_run` and `clear_dag_run` abstract methods to `base.py`
  with implementations in both `airflow_v2.py` and `airflow_v3.py`
- **CLI**: Added `delete` and `clear` subcommands to `af runs` with `--yes` and `--dry-run` flags
- **MCP tools**: Added `delete_dag_run` and `clear_dag_run` tools to `tools/dag_run.py`
  - `clear_dag_run` defaults to `dry_run=True` for safety; CLI defaults to clearing with confirmation prompt
  - `delete_dag_run` fetches run details before deleting so the agent can show what was removed
- **Tests**: 11 new unit tests covering adapters (v2/v3, 404 handling, dry_run default) and MCP tool impls;
  4 new integration tests for delete and clear endpoints against real Airflow instances
- **Docs**: Updated README, AGENTS.md, and airflow/debugging-dags skills

Note: `trigger_dag_and_wait` was also added to the README tools table — it was a pre-existing
tool that was missing from the table, not a new addition in this PR.

### DIFF
--- a/astro-airflow-mcp/README.md
+++ b/astro-airflow-mcp/README.md
@@ -475,7 +475,7 @@ To opt out:
 af telemetry disable
 ```
 
-You can also disable telemetry by setting the `AF_TRACKING_DISABLED=1` environment variable.
+You can also disable telemetry by setting the `AF_TELEMETRY_DISABLED=1` environment variable.
 
 ## Architecture
 
--- a/.cursor-plugin/marketplace.json
+++ b/.cursor-plugin/marketplace.json
@@ -1,0 +1,17 @@
+{
+  "name": "astronomer",
+  "owner": {
+    "name": "Astronomer",
+    "email": "support@astronomer.io"
+  },
+  "metadata": {
+    "description": "Data engineering plugins by Astronomer"
+  },
+  "plugins": [
+    {
+      "name": "data",
+      "source": "./",
+      "description": "Data engineering plugin - warehouse exploration, pipeline authoring, Airflow integration"
+    }
+  ]
+}
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -1,0 +1,39 @@
+{
+  "name": "data",
+  "displayName": "Astronomer Data Engineering",
+  "version": "0.1.0",
+  "description": "Data engineering plugin - warehouse exploration, pipeline authoring, Airflow integration",
+  "author": {
+    "name": "Astronomer",
+    "email": "support@astronomer.io"
+  },
+  "homepage": "https://github.com/astronomer/agents",
+  "repository": "https://github.com/astronomer/agents",
+  "license": "Apache-2.0",
+  "keywords": [
+    "data-engineering",
+    "airflow",
+    "snowflake",
+    "bigquery",
+    "jupyter",
+    "astronomer",
+    "dbt",
+    "cosmos",
+    "openlineage"
+  ],
+  "logo": "https://www.astronomer.io/logo/astronomer-dark.svg",
+  "hooks": {
+    "hooks": {
+      "beforeSubmitPrompt": [
+        {
+          "command": "./skills/airflow/hooks/airflow-skill-suggester.sh"
+        }
+      ],
+      "stop": [
+        {
+          "command": "uv run ./skills/analyzing-data/scripts/cli.py stop"
+        }
+      ]
+    }
+  }
+}
--- a/README.md
+++ b/README.md
@@ -273,10 +273,13 @@ my_warehouse:
     - RAW
 ```
 
+> [!NOTE]
+> The `account` field requires your Snowflake **account identifier** (e.g., `orgname-accountname` or `xy12345.us-east-1`), not your account name. Find this in your Snowflake console under Admin > Accounts.
+
 Store credentials in `~/.astro/agents/.env`:
 
 ```bash
-SNOWFLAKE_ACCOUNT=xyz12345
+SNOWFLAKE_ACCOUNT=myorg-myaccount  # Use your Snowflake account identifier (format: orgname-accountname or accountname.region)
 SNOWFLAKE_USER=myuser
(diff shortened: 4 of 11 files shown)


astro-airflow-mcp/src/astro_airflow_mcp/tools/dag_run.py
```diff
@@ -378,3 +378,108 @@ def trigger_dag_and_wait(
...
+    metadata will be permanently deleted.
+
+    Args:
+        dag_id: The ID of the DAG (e.g., "example_dag")
+        dag_run_id: The ID of the DAG run to delete (e.g., "manual__2024-01-01T00:00:00+00:00")
+
+    Returns:
+        JSON confirming the deletion
+    """
+    return _delete_dag_run_impl(dag_id=dag_id, dag_run_id=dag_run_id)
```
kaxil, line 463: The CLI `delete` command has a confirmation prompt. `clear_dag_run` below has `dry_run=True` as a safe default. But this MCP tool has no equivalent safety mechanism — an AI agent can permanently delete a DAG run with zero friction.

This is the only destructive, irreversible operation we expose to AI agents without any guardrail. Should we at minimum add a docstring instruction like "Always confirm with the user before calling this tool", or better yet, have the impl fetch and return the run details first so the agent can show the user what it's about to delete?

astro-airflow-mcp/src/astro_airflow_mcp/adapters/airflow_v3.py
```diff
@@ -404,6 +404,39 @@ def get_task_logs(
                 "task logs", alternative="Check if the task instance exists and has been executed"
             )
 
+    def delete_dag_run(self, dag_id: str, dag_run_id: str) -> dict[str, Any]:
```
kaxil, line 407: Both `delete_dag_run` and `clear_dag_run` are byte-for-byte identical between `airflow_v2.py` and `airflow_v3.py`. That works because the relative endpoint paths are the same and `_delete`/`_post` build the full URL from `self.api_base_path`.

But have we verified that Airflow 3's `POST .../dagRuns/{id}/clear` accepts the same request body (`{"dry_run": bool}`) and returns the same response shape as Airflow 2? And same question for `DELETE .../dagRuns/{id}`. I don't see new integration tests covering these endpoints against real Airflow instances.

astro-airflow-mcp/README.md
```diff
@@ -206,6 +207,9 @@ claude mcp add airflow -e AIRFLOW_API_URL=https://your-airflow.example.com -e AI
 | `list_dag_runs` | Get DAG run history |
 | `get_dag_run` | Get specific DAG run details |
 | `trigger_dag` | Trigger a new DAG run (start a workflow execution) |
+| `trigger_dag_and_wait` | Trigger a DAG run and wait for completion |
```
kaxil, line 210: nit: `trigger_dag_and_wait` is a pre-existing tool that was just missing from this table — not something added in this PR. Worth calling that out in the PR description so reviewers don't wonder if it's new.

astro-airflow-mcp/tests/test_adapters.py
```diff
@@ -987,6 +987,176 @@ def test_clear_task_instances_v3_handles_404(self, mocker):
...
+        result = adapter.delete_dag_run("example_dag", "manual__2024-01-01")
+
+        assert result == {}
+        mock_client.delete.assert_called_once()
+        call_args = mock_client.delete.call_args
+        assert "/api/v2/dags/example_dag/dagRuns/manual__2024-01-01" in call_args[0][0]
+
+    def test_delete_dag_run_handles_404(self, mocker):
+        """Test delete_dag_run raises NotFoundError on 404."""
+        adapter = AirflowV3Adapter(
```
kaxil, line 1049: Minor: this test creates the adapter without `token_getter`, while the success tests above (`test_delete_dag_run_v2`, `test_delete_dag_run_v3`) provide one. The inconsistency could mask auth-related issues in error paths. Would keep it consistent.

astro-airflow-mcp/src/astro_airflow_mcp/cli/runs.py
```diff
@@ -215,6 +215,65 @@ def _get_failed_task_instances(
...
+    try:
+        adapter = get_adapter()
+        adapter.delete_dag_run(dag_id, dag_run_id)
+        output_json({"message": f"DAG run '{dag_run_id}' deleted", "dag_id": dag_id})
+    except Exception as e:
+        output_error(str(e))
+
+
+@app.command("clear")
+def clear_dag_run(
```
kaxil, line 243: The CLI defaults to actually clearing (with confirmation prompt), while the MCP tool defaults to `dry_run=True`. That's a reasonable split — CLI users are deliberate, AI agents should be cautious — but it's not documented anywhere. A short note in the `--dry-run` help text or the docstring would save someone the confusion later.

astro-airflow-mcp/src/astro_airflow_mcp/tools/dag_run.py
tayloramurphy: @kaxil so I updated this but I have a meta question here - should an agent-first CLI be prompting the user? The whole point with the af cli is to be agent-first and have limited interactivity. Maybe then we just shouldn't be doing destructive things via the CLI? That said, the agent can also just call the `api` directly...

kaxil: iirc if --skip-dangerously-permission flag is set it will ignore it. Good point about agent can just call api though

PR #138, opened 2026-02-24
Implement AF_READ_ONLY Mode to Block Write Operations

igorastronomer:
This PR introduces a read-only safety guard for `astro-airflow-mcp`. It allows users and agents to safely interact with Airflow environments by restricting the client to read-only (`GET`) requests.

## Key Changes
* **Environment Variable Guard:** When `AF_READ_ONLY` is set to 'true', any state-mutating API calls (`POST`, `PATCH`, `PUT`, `DELETE`) are blocked automatically at the adapter request layer.
* **Informative Errors:** Attempted writes trigger a ReadOnlyError, which returns a clear message detailing the blocked operation and active environment variable to the user/agent.
* **Guarded Toolset:** Core actions like `pause_dag`, `unpause_dag`, `trigger_dag_run`, and `clear_task_instances` successfully respect this flag and will surface the rejection context without executing against Airflow.
* **Testing:** Added robust unit and E2E coverage in `tests/test_read_only.py` to ensure read commands (`list_dags`, etc.) still succeed, and all writes are properly intercepted and verified.

### DIFF
--- a/astro-airflow-mcp/tests/test_read_only.py
+++ b/astro-airflow-mcp/tests/test_read_only.py
@@ -5,6 +5,7 @@ while read operations continue to work normally.
 """
 
 import json
+from contextlib import nullcontext as does_not_raise
 from unittest.mock import MagicMock
 
 import pytest
@@ -17,7 +18,8 @@ class TestAssertWritable:
 
     def test_allowed_when_env_not_set(self, monkeypatch):
         monkeypatch.delenv("AF_READ_ONLY", raising=False)
-        _assert_writable("POST dags/test/dagRuns")  # should not raise
+        with does_not_raise():
+            _assert_writable("POST dags/test/dagRuns")
 
     @pytest.mark.parametrize("value", ["true", "True", "TRUE", " true ", "TRUE "])
     def test_blocked_with_true_values(self, monkeypatch, value):
@@ -28,7 +30,8 @@ class TestAssertWritable:
     @pytest.mark.parametrize("value", ["0", "1", "false", "no", "yes", "", "off"])
     def test_allowed_with_non_true_values(self, monkeypatch, value):
         monkeypatch.setenv("AF_READ_ONLY", value)
-        _assert_writable("POST dags/test/dagRuns")  # should not raise
+        with does_not_raise():
+            _assert_writable("POST dags/test/dagRuns")
 
     def test_error_message_includes_operation_and_env_var(self, monkeypatch):
         monkeypatch.setenv("AF_READ_ONLY", "true")


astro-airflow-mcp/README.md
```diff
@@ -149,6 +149,7 @@ By default, the server connects to `http://localhost:8080` (Astro CLI default).
 | `AIRFLOW_AUTH_TOKEN` | Bearer token (alternative to username/password) |
 | `AIRFLOW_VERIFY_SSL` | Set to `false` to disable SSL certificate verification |
 | `AIRFLOW_CA_CERT` | Path to custom CA certificate bundle |
+| `AF_READ_ONLY` | Set to `1`, `true`, or `yes` to block all write operations |
```
josh-fell: Is the presumption this would only be set by humans? If so, would it be worth having a single, explicit value in the documentation e.g., `true`, rather than a few truthy options?

igorastronomer: That's a good point - will update that

astro-airflow-mcp/src/astro_airflow_mcp/adapters/base.py
```diff
@@ -1,11 +1,30 @@
...
+    def __init__(self, operation: str):
+        super().__init__(
+            f"Operation '{operation}' is blocked: running in read-only mode "
+            f"(${READ_ONLY_ENV_VAR}=1). Only read operations are allowed."
+        )
+
+
+def _assert_writable(operation: str) -> None:
+    """Raise ReadOnlyError if AF_READ_ONLY is set to a truthy value."""
+    if os.environ.get(READ_ONLY_ENV_VAR, "").strip().lower() in ("1", "true", "yes"):
```
josh-fell:
```suggestion
    if os.environ.get(READ_ONLY_ENV_VAR, "").strip().lower() == "true":
```
Since we're explicitly stating what the value should be, we should refine this too -- although seems reasonable to implicitly handle True, TRUE, true, etc. patterns.

astro-airflow-mcp/tests/test_read_only.py
```diff
@@ -0,0 +1,141 @@
...
+    @pytest.mark.parametrize("value", ["true", "True", "TRUE", " true ", "TRUE "])
+    def test_blocked_with_true_values(self, monkeypatch, value):
+        monkeypatch.setenv("AF_READ_ONLY", value)
+        with pytest.raises(ReadOnlyError, match="read-only mode"):
+            _assert_writable("POST dags/test/dagRuns")
+
+    @pytest.mark.parametrize("value", ["0", "1", "false", "no", "yes", "", "off"])
+    def test_allowed_with_non_true_values(self, monkeypatch, value):
+        monkeypatch.setenv("AF_READ_ONLY", value)
+        _assert_writable("POST dags/test/dagRuns")  # should not raise
```
josh-fell: Could you add a assertion to the tests where there is no raised exception?

You could use `contextlib.nullcontext` within `pytest.raises()` for this type of thing; something like:
```py
from contextlib import nullcontext as does_not_raise

@pytest.mark.parametrize("value", ["0", "1", "false", "no", "yes", "", "off"])
    def test_allowed_with_non_true_values(self, monkeypatch, value):
        monkeypatch.setenv("AF_READ_ONLY", value)
        with pytest.raises(does_not_raise()):
            _assert_writable("POST dags/test/dagRuns") 
```

josh-fell (approved):
Just a comment to make sure there are assertions in some of the tests, but LGTM

PR #141, opened 2026-02-27
Improve agent detection

schnie:
Improves agent detection and fixes a few unrelated build issues.


astro-airflow-mcp/src/astro_airflow_mcp/telemetry.py
```diff
@@ -158,6 +171,7 @@ def _send(api_url: str, body: dict, *, debug: bool = False) -> None:
             stderr=None if debug else subprocess.DEVNULL,
             start_new_session=not debug,
         )
+        assert proc.stdin is not None
```
josh-fell: Just curious, was this to make type-checking happy? If so, should we wrap this behind `TYPE_CHECKING`?

schnie: Ah good catch - it was just there to ensure the following lines (`write()`, `close()`) would work. Replaced with a cleaner `if` statement.

PR #154, opened 2026-03-04
Add airflow-plugins skill

vojay-dev:
Adds a skill for building Airflow 3.1+ plugins with FastAPI, covering plugin registration, JWT auth, asyncio patterns, static serving, macros, middleware, operator links, and the full DAG/run/task API surface.

### DIFF
--- a/skills/airflow-plugins/SKILL.md
+++ b/skills/airflow-plugins/SKILL.md
@@ -275,7 +275,9 @@ The pattern is always: define a plain inner `def _fetch()` with all SDK logic, t
 
 ### Alternative: Direct database access
 
-Since FastAPI plugin endpoints run inside the **API server process** (not in a task worker), they have direct access to Airflow's internal SQLAlchemy models — no HTTP round-trip or JWT needed:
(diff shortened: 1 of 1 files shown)


skills/airflow-plugins/SKILL.md
```diff
@@ -0,0 +1,555 @@
...
+                return DagApi(client).get_dags(limit=100).dags
+        dags = await asyncio.to_thread(_fetch)
+        return [{"dag_id": d.dag_id, "is_paused": d.is_paused} for d in dags]
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
+```
+
+The pattern is always: define a plain inner `def _fetch()` with all SDK logic, then `await asyncio.to_thread(_fetch)`.
+
+### Alternative: Direct database access
```
TJaniF, line 276: I was about to say we maybe should not tell the Agents about this, but they will try it anyways so I think it makes sense to have it here but I maybe would put the warning further up that this interface is not supposed to be public and obv you can cause issues in the metadb so any use of this pattern needs to be flagged to the user with that warning.

TJaniF (approved):
Works great! :)

PR #155, opened 2026-03-04
Add MCP App visualization server for analyzing-data skill

kaxil:
## Summary

- Adds `viz_mcp.py` — a Prefab-based MCP App server with `render_chart` and `render_table` tools for inline interactive visualizations (bar/line charts, sortable/searchable/paginated tables with tab switching)
- Registers `analytics-viz` MCP server in `.cursor-plugin/plugin.json` (MCP Apps are Cursor-only)
- Updates `analyzing-data/SKILL.md` with visualization workflow guidance so the LLM knows when and how to call the viz tools after running SQL

image

### DIFF
--- a/.cursor-plugin/plugin.json
+++ b/.cursor-plugin/plugin.json
@@ -36,5 +36,15 @@
         }
       ]
     }
+  },
+  "mcpServers": {
+    "analytics-viz": {
+      "command": "uv",
+      "args": [
+        "run",
+        "./skills/analyzing-data/scripts/viz_mcp.py",
+        "--stdio"
+      ]
+    }
   }
 }
--- a/skills/analyzing-data/SKILL.md
+++ b/skills/analyzing-data/SKILL.md
@@ -42,6 +42,10 @@ Answer business questions by querying the data warehouse. The kernel auto-starts
    ```
 
 6. **Present findings** to user.
+7. **Optional visualization** — If the user asks for chart/table visuals and the `analytics-viz` MCP server is available:
+   - Use `render_table(rows=[...], title="...")` for tabular output
+   - Use `render_chart(rows=[...], x_key="...", y_key="...", chart_type="line|bar", title="...", top_n=<int>, sort_desc=<bool>)` for trend/comparison views
+   - The UI has controls to filter/sort/switch chart type within loaded data. Users can ask in chat for a different row count.
 
 ## Kernel Functions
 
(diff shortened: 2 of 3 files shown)


.cursor-plugin/plugin.json
josh-fell, line 1: You mention a Claude plugin too in the PR description. Do you have that update locally to push?

skills/analyzing-data/scripts/viz_mcp.py
```diff
@@ -0,0 +1,225 @@
...
+            sort_desc=sort_desc,
+        ),
+    )
+
+
+if __name__ == "__main__":
+    if "--stdio" in sys.argv:
+        mcp.run(transport="stdio")
+    else:
+        mcp.run(transport="stdio")
```
josh-fell:
```suggestion
if __name__ == "__main__":
    mcp.run(transport="stdio")
```
Or does the conditional handle some nuance in the plugins

skills/analyzing-data/SKILL.md
```diff
@@ -42,6 +42,10 @@ Answer business questions by querying the data warehouse. The kernel auto-starts
    ```
 
 6. **Present findings** to user.
+7. **Optional visualization** — If the user asks for chart/table visuals and the `analytics-viz` MCP server is available:
```
josh-fell: Do we need to add color for the agent to determine whether if the MCP server is available?

.cursor-plugin/plugin.json
kaxil: Already done — removed the `mcpServers` block from `.claude-plugin/plugin.json` since MCP Apps are Cursor-only. PR description updated too.

skills/analyzing-data/scripts/viz_mcp.py
kaxil: Good catch — no nuance, both branches were identical. Simplified to just `mcp.run(transport="stdio")` in d3206fe.

skills/analyzing-data/SKILL.md
kaxil: Reworded to "if `render_table`/`render_chart` tools are in your tool list" — the agent sees its available tools directly, so no extra heuristic needed. Fixed in d3206fe.

.cursor-plugin/plugin.json
kaxil: Claude Code doesn't supprot MCP Apps unfortunately. And Claude CoWork is sandboxed by default to allow a lot of what we use

josh-fell (approved):

PR #156, opened 2026-03-05
Make discovery token names user-specific to avoid collisions

### LINKED ISSUE #153: `af instances discover` astro backend tries to create tokens with the same name
We have multiple users trying to use the Airflow skills and MCP with Claude. The easiest way to set this up is with the handy `af instances discover` command. However, for Astro instances, the name of the token it generates is hardcoded [here](https://github.com/astronomer/agents/blob/main/astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py#L75). So whenever someone else runs this command it fails because they don't already have the token. It would be great if these tokens could be uniquely named OR if we can just reuse the token from the Astro CLI (much prefer this second option as opposed to having a bunch of adhoc tokens lying around).

tayloramurphy:
## Summary

Fixes #153

- Discovery token names were hardcoded as `af-cli-discover`, causing collisions when multiple users ran `af instances discover` on the same deployment
- Token names are now user-specific (e.g., `af-discover-taylor-murphy`) using the Astro CLI email from `~/.astro/config.yaml`, with OS username as fallback
- Token name is resolved once at backend init and reused for the session

## Test plan

- [x] Existing tests updated and passing (85 total)
- [x] Run `af instances discover astro --dry-run` and verify token name includes your username

### DIFF
--- a/astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import getpass
+import os
 import re
 import shutil
 import subprocess  # nosec B404 - subprocess is needed for CLI wrapper
@@ -108,7 +109,10 @@ class AstroCli:
         email = self._get_user_email()
         if email:
             return email.split("@")[0]
-        return getpass.getuser()
+        try:
+            return getpass.getuser()
+        except KeyError:
+            return "unknown"
 
     def _get_user_email(self) -> str | None:
         """Get the user's email from Astro CLI config.
@@ -116,7 +120,8 @@ class AstroCli:
         Returns:
             Email string or None if unavailable
         """
-        config_path = Path.home() / ".astro" / "config.yaml"
+        astro_home = os.environ.get("ASTRO_HOME", Path.home() / ".astro")
+        config_path = Path(astro_home) / "config.yaml"
         if not config_path.exists():
             return None
 
--- a/astro-airflow-mcp/tests/test_astro_cli.py
+++ b/astro-airflow-mcp/tests/test_astro_cli.py
@@ -534,6 +534,34 @@ class TestTokenName:
         with patch("astro_airflow_mcp.discovery.astro_cli.Path.home", return_value=tmp_path):
             assert cli._get_user_email() is None
 
+    def test_get_token_name_falls_back_to_unknown_when_getuser_fails(self):
+        """Test token name falls back to 'unknown' when getpass.getuser() raises KeyError."""
+        cli = AstroCli()
+        with (
+            patch.object(cli, "_get_user_email", return_value=None),
+            patch(
+                "astro_airflow_mcp.discovery.astro_cli.getpass.getuser",
+                side_effect=KeyError("no user"),
+            ),
+        ):
+            assert cli.get_token_name() == "af-discover-unknown"
+
+    def test_get_user_email_respects_astro_home(self, tmp_path):
+        """Test _get_user_email uses ASTRO_HOME env var when set."""
+        custom_astro_home = tmp_path / "custom_astro"
+        custom_astro_home.mkdir()
+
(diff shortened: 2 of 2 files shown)


astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro.py
```diff
@@ -49,15 +49,19 @@ class AstroDiscoveryBackend:
...
         Args:
             cli: Optional AstroCli instance (creates one if not provided)
         """
         self._cli = cli or AstroCli()
+        self._token_name = self._cli.get_token_name()
+
+    @property
+    def token_name(self) -> str:
+        """Get the user-specific token name for discovery."""
+        return self._token_name
```
kaxil, line 69: No migration path for existing `af-cli-discover` tokens. Users who already ran discovery will have tokens with the old name — after this change `token_exists()` checks the new name, so discovery will create a **second** token on the same deployment and the old one becomes orphaned. Worth at least a log/warning when the legacy name is found, or a note in the README telling users to clean up.

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py
```diff
@@ -72,12 +74,70 @@ class AstroCli:
...
+            return f"{self.TOKEN_NAME_PREFIX}-{normalized}"
+        return self.TOKEN_NAME_PREFIX
+
+    def _get_user_identifier(self) -> str:
+        """Get a deterministic user identifier.
+
+        Tries user_email from ~/.astro/config.yaml first (email local part),
+        then falls back to the OS username.
+
+        Returns:
```
kaxil, line 106: `getpass.getuser()` can raise `KeyError` on environments where `LOGNAME`/`USER`/`USERNAME` are unset and there is no pwd entry. Since this runs at `AstroDiscoveryBackend.__init__` time, an unhandled exception here would break construction. Wrapping in try/except with a fallback (e.g. `"unknown"`) would be safer.

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py
```diff
@@ -72,12 +74,70 @@ class AstroCli:
...
+
+        Generates a deterministic token name like 'af-discover-firstname-lastname'
+        using the user's Astro email, or OS username as fallback. This avoids
+        collisions when multiple users run discovery on the same deployment.
+
+        Returns:
+            Token name string (e.g., 'af-discover-firstname-lastname')
+        """
+        identifier = self._get_user_identifier()
+        normalized = re.sub(r"[^a-z0-9]+", "-", identifier.lower()).strip("-")
```
kaxil, line 95: Hardcodes `~/.astro/config.yaml` — this bypasses any `ASTRO_HOME` override the user might have set. Minor, but worth noting since the rest of the class delegates to the CLI binary which respects that env var.

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro.py
kaxil: cc @schnie if you have any suggestions

schnie: Yea I suppose actually deleting the old named token might not be advisable either since someone else on the team may be using it still 🤔 . Yea maybe just a warning message if the old naming convention is found in their astro tokens.

schnie: Added a warning message.

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py
schnie: Fixed.

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py
schnie: Fixed.

PR #185, opened 2026-04-15
Add ruff-airflow-rule skill for creating AIR lint rules

Dev-iL:
The new SKILL guides contributors through the full workflow of adding a new Airflow-specific lint rule to Ruff: rule selection, code authoring, registration, test fixtures, snapshot acceptance, and validation against a local Airflow repo.

Covers both general best-practice rules (AIR0xx, AIR2xx) and Airflow 2→3 migration rules (AIR301–AIR321), with templates, shared-helper references, documentation conventions, and dual-dispatch patterns for TaskFlow and operator APIs.

### DIFF
--- a/skills/ruff-airflow-rule/SKILL.md
+++ b/skills/ruff-airflow-rule/SKILL.md
@@ -1,0 +1,897 @@
+---
+name: ruff-airflow-rule
+description: This skill should be used when the user asks to "add a new airflow rule", "create an airflow lint rule", "implement an airflow inspection", "new AIR rule", or discusses creating Ruff linter rules in the airflow category.
+---
+
+# Creating a New Airflow Lint Rule in Ruff
+
+This skill guides creating new Airflow-specific lint rules (AIR prefix) in the Ruff codebase.
+
+## Pre-Implementation: Gather Context
+
+**Before writing any code, ask the user the following questions (if not already answered in their request):**
+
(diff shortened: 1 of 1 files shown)


kaxil (commented):
Apologies for the delay, I thought about this, I think this should stay in Airflow repo instead of https://github.com/astronomer/agents -- since it is more for developers than for end users

Dev-iL:
I see what you're saying, and it's true that this skill is not "the problem" of this repo.

Here's the issue I see with your suggestion: a skill like this works against the Ruff repo. The Airflow repo is not currently set up to be installable as a plugin, so including this skill there will create a non-standard multi-repo workflow where the user runs an agent in Ruff but tells it to use a skill from Airflow, or runs the agent in Airflow and tell it to modify another repo. Doing it this way will result in a bad developer experience and likely cause the skill to be undiscoverable and unused.

The options I see are:
1. Add a marketplace to Airflow and put this skill there.
2. Add it under the Ruff repo - a direction that's being explore here: https://github.com/astral-sh/ruff/discussions/24812
3. Create a dedicated plugin for it in a separate repo. An overkill but is the easiest solution where it comes to community approvals.

kaxil:
2. Is a good option

Even without 2, if it is in Airflow repo, the only thing that needs to be done, is `/add-dir /path/to/ruff/repo`

PR #187, opened 2026-04-18
add dag-factory skill

gyli:
Skill proposed here: https://github.com/astronomer/dag-factory/issues/687

### DIFF
--- a/skills/dag-factory/SKILL.md
+++ b/skills/dag-factory/SKILL.md
@@ -10,6 +10,7 @@ You are helping a user build Apache Airflow DAGs declaratively with **dag-factor
 > **Package**: `dag-factory` on PyPI
 > **Repo**: https://github.com/astronomer/dag-factory
 > **Docs**: https://astronomer.github.io/dag-factory/latest/
+> **Targets**: dag-factory **v1.0+** only. For pre-1.0 projects, see [reference/migration.md](reference/migration.md) before applying any guidance from this skill.
 > **Requires**: Python 3.10+, Airflow 2.4+ (Airflow 3 supported)
 
 ## Before Starting
@@ -17,7 +18,7 @@ You are helping a user build Apache Airflow DAGs declaratively with **dag-factor
 Confirm with the user:
 1. **Airflow version** ≥2.4
 2. **Python version** ≥3.10
-3. **dag-factory version**: v1.0+ recommended. If migrating from <1.0, see **Migration Notes** — there are several breaking changes.
+3. **dag-factory version**: this skill targets **v1.0+**. If the project is on <1.0, follow [reference/migration.md](reference/migration.md) to upgrade before continuing.
 4. **Use case**: dag-factory is for declarative, low-code DAG authoring. If the user needs reusable, validated Pythonic templates with Pydantic, suggest **blueprint** instead. If they need full Python flexibility, suggest the **authoring-dags** skill.
 
 ---
@@ -36,7 +37,7 @@ Confirm with the user:
 | "Use a timetable" / "datetime in YAML" / "timedelta in YAML" | Go to **Custom Python Objects (`__type__`)** |
 | "Lint my YAML" / "Validate" | Go to **Validation Commands** |
 | "Convert Airflow 2 YAML to Airflow 3" | Go to **Validation Commands** (`dagfactory convert`) |
-| "Migrate from dag-factory <1.0" | Go to **Migration Notes** |
+| "Migrate from dag-factory <1.0" | See [reference/migration.md](reference/migration.md) |
 | dag-factory errors / troubleshooting | Go to **Troubleshooting** |
 
 ---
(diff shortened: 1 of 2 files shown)


skills/dag-factory/SKILL.md
```diff
@@ -0,0 +1,500 @@
...
+---
+
+## Defaults
+
+There are four ways to set defaults, in **precedence order** (highest first):
+
+1. `default_args` / DAG-level keys inside an individual DAG
+2. The top-level `default:` block in the same YAML file
+3. `defaults_config_dict=` argument to `load_yaml_dags`
+4. A `defaults.yml` (or `defaults.yaml`) file via `default_args_config_path=` (or auto-detected next to the DAG YAML)
```
kaxil: Small but load-bearing: the actual function parameter in `dagfactory/dagfactory.py` is `defaults_config_path`, not `default_args_config_path` (verified in v1.0.1: `__init__` on line 45, `load_yaml_dags` on line 280). The old name exists only as an internal attribute (`self.default_args_config_path`). Same mismatch on lines 186, 471, 473.

skills/dag-factory/SKILL.md
```diff
@@ -0,0 +1,500 @@
...
+      __type__: kubernetes.client.models.V1ResourceRequirements
+      limits: {cpu: "1", memory: "1024Mi"}
+      requests: {cpu: "0.5", memory: "512Mi"}
+```
+
+---
+
+## Dynamic Task Mapping
+
+Use `expand` and `partial` keys on a task to map dynamically. Reference the output of another task with the `+` prefix or by name (`task_id.output` works for some forms; the `+` syntax is dag-factory's idiomatic resolver).
```
kaxil: This paragraph says `task_id.output` "works for some forms" and `+` is idiomatic. In the example right below, `op_args: request.output` has no `+`, which makes the guidance contradict the example. Worth either showing both forms explicitly or picking one: `+name` is used as the bare value reference (`expand: {number: +numbers_list}`), while `task_id.output` is the XCom-style form used inside `op_args` / `op_kwargs`.

skills/dag-factory/SKILL.md
```diff
@@ -0,0 +1,500 @@
...
+      operator: airflow.operators.bash.BashOperator
+      bash_command: "echo 'consumer'"
+```
+
+### Conditional Dataset Scheduling (Airflow 2.9+ / dag-factory 0.22+)
+
+Combine datasets with `__and__` / `__or__`:
+
+```yaml
+schedule:
```
kaxil, line 320: I think this example won't schedule. The working dag-factory example nests the operators under a `datasets:` key:

```yaml
schedule:
  datasets:
    __or__:
      - __and__: [...]
```

Without `datasets:`, the parser treats `__or__` as a top-level schedule value rather than a dataset condition tree.

skills/dag-factory/SKILL.md
```diff
@@ -0,0 +1,500 @@
...
+  __type__: airflow.timetables.trigger.CronTriggerTimetable
+  cron: "0 1 * * 3"
+  timezone: UTC
+```
+
+- `__type__` is the **full import path** to the class
+- `__args__` is a list of positional arguments
+- Other keys become keyword arguments
+- For lists of typed objects, use `__type__: builtins.list` with an `items:` key
```
kaxil, line 407: Minor: the example uses a plain `items:` key for list elements under a typed construction. The "Reserved Keys" list on line 394 doesn't include it. If `items` is reserved under `__type__: builtins.list`, worth calling that out so users don't put their own field called `items` in a typed construction.

skills/dag-factory/SKILL.md
```diff
@@ -0,0 +1,500 @@
...
+# 2. Have Airflow parse to catch operator/import errors
+#    (Astro CLI users)
+astro dev parse
+```
+
+`dagfactory lint` only checks YAML syntax — operator import errors and missing kwargs surface at Airflow parse time.
+
+---
+
+## Migration Notes (pre-1.0 → 1.0)
```
kaxil: Skill-authoring note: the authoring guide recommends avoiding version-pinned content since it goes stale. Once v1.1 lands, this entire section, plus the inline `v1.0 renamed...` callouts on lines 154, 308, 390, and the row in the migration table on line 432, becomes historical noise. Consider moving migration details into `references/migration.md` and leaving a single pointer in SKILL.md so the main flow stays version-neutral.

If you keep the migration content inline, two factual fixes are needed:
- Line 154: both kwargs were renamed in v1.0, not just the dict. `default_args_config_dict` became `defaults_config_dict` *and* `default_args_config_path` became `defaults_config_path`.
- Line 432: the migration table is missing a row for the path kwarg, `default_args_config_path=` to `defaults_config_path=`.

skills/dag-factory/SKILL.md
```diff
@@ -0,0 +1,500 @@
...
+
+### Conditional dataset schedule ignored
+
+**Cause**: Airflow <2.9, dag-factory <0.22, or using legacy `!and`/`!or` keys.
+
+**Fix**: Upgrade and rename to `__and__` / `__or__`.
+
+### Multiple `defaults.yml` not merging as expected
+
+**Cause**: `default_args_config_path` not pointing at a parent directory of the DAG YAML.
```
kaxil: Same param name issue: should be `defaults_config_path`.

skills/dag-factory/SKILL.md
gyli: Good catch! Outdated doc in dag-factory as well. I will fix in both places.

skills/dag-factory/SKILL.md
gyli: You are right, will fix

skills/dag-factory/SKILL.md
gyli: Will call out this conditional reserved keys in dag-factory doc as well: https://github.com/astronomer/dag-factory/blob/cd43250a9a05852f14bc939050248a4c1c917ad8/docs/configuration/configuring_workflows.md?plain=1#L42

skills/dag-factory/SKILL.md
gyli: Both example added.

gyli:
Hi @kaxil , thanks for the careful review! All comment addressed, please check again.

kaxil (approved):

kaxil:
Thanks @gyli

PR #190, opened 2026-04-21
Detect paused DAG in `runs trigger-wait`

jlaneve:
`af runs trigger-wait` on a paused DAG used to silently block until timeout — the trigger request succeeded but no run was ever scheduled.

The MCP `trigger_dag_and_wait` tool already auto-unpaused before triggering; this mirrors that behavior in the CLI (`runs trigger` and `runs trigger-wait`). Added `--no-auto-unpause` for callers who'd rather fail fast with a clear error than have the DAG state mutated.


schnie (approved):
love it

PR #191, opened 2026-04-21
af CLI: treat empty AIRFLOW_API_URL as explicit no-op

### LINKED ISSUE #188: af CLI: provide a way to disable the default localhost:8080 fallback
## Problem

When `AIRFLOW_API_URL` is unset and `~/.af/config.yaml` has no resolvable `current-instance`, the `af` CLI silently falls back to `DEFAULT_AIRFLOW_URL` (`http://localhost:8080`) — see `astro_airflow_mcp/cli/context.py:76-82` and `astro_airflow_mcp/constants.py`.

This is fine for interactive use against a single local Airflow, but it creates a correctness hazard for any caller that can legitimately be in a "no Airflow is configured right now" state:

- If **nothing** is on `localhost:8080`, the caller gets a confusing connection error against a URL they never asked for.
- If **something else** is on `localhost:8080` (a different project, an unrelated service answering the health probe, a developer's stale container, etc.), `af` will happily query it and return results that don't belong to the user's current context. There's no way to tell from the CLI's output that the URL came from a default rather than explicit configuration.

There is also no way to distinguish "env var intentionally empty" from "env var not set":

```python
url_from_env = bool(os.environ.get(\"AIRFLOW_API_URL\"))
```

`bool(\"\")` is `False`, so setting `AIRFLOW_API_URL=\"\"` as a "no instance" signal is silently ignored and the same default-fallback path runs.

## Desired behavior

Callers need an explicit way to say "there is no Airflow configured here — do not guess". Any of the following would solve it; they're listed roughly in order of invasiveness:

1. **Treat env-var-present-but-empty as an intentional no-op.** Change the presence check from `bool(os.environ.get(\"AIRFLOW_API_URL\"))` to `\"AIRFLOW_API_URL\" in os.environ`. An explicit empty string then means "no instance configured" and `af` exits with a clear error instead of defaulting.
2. **Add a dedicated opt-out env var** (e.g. `AF_NO_DEFAULT=1`) that suppresses the `DEFAULT_AIRFLOW_URL` fallback when both env and config resolution yield nothing.
3. **Require explicit configuration by default.** If neither the env var nor the config file yields a URL, error out with a message pointing the user at `af instance add` / `AIRFLOW_API_URL`, rather than silently defaulting. The current default could move behind an opt-in flag for truly interactive first-time use.

Option 1 is the smallest change and is backward-compatible (no existing caller sets `AIRFLOW_API_URL=\"\"`, so nothing breaks). Option 3 is the most correct long-term but is a behavior change for first-run users.

## Why it matters

Programmatic callers (automation, orchestration wrappers, agent frameworks) can reliably detect whether the current working environment has a configured Airflow, and need a way to propagate "nothing is configured" to `af` without risking a misleading result from whatever happens to be listening on `localhost:8080`. Today they cannot.

## Acceptance

- A caller can set a single env var (or not set any env var, per option 3) such that `af config version` / `af dags list` / etc. fail with a clear "no Airflow configured" error instead of querying `DEFAULT_AIRFLOW_URL`.
- Documented in `README.md` or the CLI `--help` so integrators know the contract.

jlaneve:
Setting `AIRFLOW_API_URL=""` now exits with a clear error instead of silently falling back to `http://localhost:8080`. This lets programmatic callers (automation, agent frameworks) signal "no Airflow is configured right now" without risking queries against whatever happens to be listening on localhost.

Unset `AIRFLOW_API_URL` still falls back to config → default, so existing behavior is preserved. Documented in `--help` and the README.

Fixes https://github.com/astronomer/agents/issues/188

### DIFF
--- a/astro-airflow-mcp/src/astro_airflow_mcp/cli/context.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/cli/context.py
@@ -75,27 +75,26 @@ class CLIContext:
         # since the config's auth is for a different instance.
         #
         # An explicitly empty AIRFLOW_API_URL ("" in the environment) is treated
-        # as "no Airflow is configured right now" — do NOT fall back to the
+        # as "no Airflow is configured right now" -- do NOT fall back to the
         # config file or DEFAULT_AIRFLOW_URL. This lets programmatic callers
         # (automation, agent frameworks) propagate "nothing is configured"
         # without risking queries against whatever happens to be listening on
         # localhost:8080.
-        url_env_value = os.environ.get("AIRFLOW_API_URL")
-        url_env_present = "AIRFLOW_API_URL" in os.environ
-        url_from_env = bool(url_env_value)
+        airflow_url_env = os.environ.get("AIRFLOW_API_URL")
+        url_from_env = bool(airflow_url_env)
 
-        if url_env_present and not url_env_value:
-            print(
-                "Error: AIRFLOW_API_URL is set but empty. No Airflow instance "
-                "is configured.\n"
+        if airflow_url_env is not None and not airflow_url_env:
+            message = (
+                "Error: AIRFLOW_API_URL is set but empty. "
+                "No Airflow instance is configured.\n"
                 "To configure one, either:\n"
                 "  - set AIRFLOW_API_URL to a webserver URL, or\n"
-                "  - run `af instance add <name> --url <url>` and "
-                "`af instance use <name>`.\n"
-                "To use the default (http://localhost:8080), unset "
-                "AIRFLOW_API_URL.",
-                file=sys.stderr,
+                "  - run `af instance add <name> --url <url>` "
+                "and `af instance use <name>`.\n"
+                "To use the default (http://localhost:8080), "
+                "unset AIRFLOW_API_URL."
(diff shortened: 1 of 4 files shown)


astro-airflow-mcp/src/astro_airflow_mcp/cli/context.py
```diff
@@ -73,7 +73,30 @@ def init(self) -> None:
...
         # since the config's auth is for a different instance.
-        url_from_env = bool(os.environ.get("AIRFLOW_API_URL"))
+        #
+        # An explicitly empty AIRFLOW_API_URL ("" in the environment) is treated
+        # as "no Airflow is configured right now" -- do NOT fall back to the
+        # config file or DEFAULT_AIRFLOW_URL. This lets programmatic callers
+        # (automation, agent frameworks) propagate "nothing is configured"
+        # without risking queries against whatever happens to be listening on
+        # localhost:8080.
+        airflow_url_env = os.environ.get("AIRFLOW_API_URL")
```
kaxil, line 83: I think @schnie added this for working with local instances without configuring anything. @schnie could you review this please?

schnie (approved):
Makes sense

PR #192, opened 2026-04-27
Fix list_dag_runs (MCP tool + af CLI) returning oldest runs first

### LINKED ISSUE #168: af runs list: show most recent runs first by default
## Problem

`af runs list -d <dag_id>` returns runs in ascending order by default (oldest first). This means the most recent runs appear at the bottom, making it easy to miss new activity when debugging.

## Expected behavior

Results should be sorted by `start_date` descending (most recent first) by default, consistent with how most users want to investigate DAG runs.

## Workaround

Currently requires piping or manually scrolling to the end of the output.

JoaVirtudes19:
The `list_dag_runs` MCP tool only accepted `dag_id` and called Airflow's `GET /dags/{dag_id}/dagRuns` with no `order_by`. Airflow's default sort on that endpoint is `id ASC`, so for any DAG with more than 100 runs the tool returned the oldest 100 runs and the recent ones were unreachable through the MCP — directly contradicting the docstring's promise of "sorted by most recent".

The `af runs list` CLI hit the same root cause: `--order-by` defaulted to `None`, so the API again returned oldest-first. Issue #168 reported this from the CLI side.

## Changes

**MCP tool (`list_dag_runs`)**
- Exposes `limit`, `offset`, and `order_by` on the tool surface
- Defaults `order_by="-start_date"` so callers get newest-first

**CLI (`af runs list`)**
- Changes `--order-by` default from `None` to `-start_date`, matching the MCP tool

## Why `-start_date`

`start_date` is the only sort field available on both Airflow 2.x and 3.x DAG-run list endpoints. `run_after` is Airflow 3 only; `logical_date` doesn't exist in v2 (it's `execution_date` there). Picking `-start_date` keeps a single default that works across both adapters.

Callers who want a different order (e.g. `id` ascending to match the old Airflow default, or filter by state) can still pass `--order-by` explicitly on the CLI or `order_by=` on the MCP tool.

Closes #168

### DIFF
--- a/astro-airflow-mcp/src/astro_airflow_mcp/cli/runs.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/cli/runs.py
@@ -29,9 +29,15 @@ def list_dag_runs(
         typer.Option("--offset", "-o", help="Offset for pagination"),
     ] = 0,
     order_by: Annotated[
-        str | None,
-        typer.Option("--order-by", help="Sort field (prefix - for descending, e.g., -start_date)"),
-    ] = None,
+        str,
+        typer.Option(
+            "--order-by",
+            help=(
+                "Sort field; prefix with '-' for descending. Defaults to '-start_date' "
+                "so the most recent runs come first."
+            ),
+        ),
+    ] = "-start_date",
     state: Annotated[
         str | None,
         typer.Option("--state", "-s", help="Filter by state: running, success, failed, queued"),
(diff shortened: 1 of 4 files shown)


kaxil (approved):

kaxil:
Thanks for the PR @JoaVirtudes19 -- I pushed a couple more changes so I can release it today

PR #204, opened 2026-04-30
af: discover Astro deployments with PAT auth instead of minting tokens

jlaneve:
## summary

`af instance discover astro` now emits instances configured for `astro_pat` auth: af reuses the user's `astro login` session and refreshes it via Auth0 directly when the token is near expiry. The old behavior, minting a permanent `DEPLOYMENT_ADMIN` token per deployment via `astro deployment token create`, is gone. Those tokens persisted in Astro until manually revoked, weren't tied to user identity, and accumulated across discovery runs.

## what changed

- **astro_pat module (new)**: `AstroPATResolver` reads the active context's bearer + refresh token from `~/.astro/config.yaml`, refreshes via Auth0's `oauth/token` endpoint when the cached token is within 60s of expiry, serializes refreshes via a thread lock, and honors `ASTRO_API_TOKEN` for headless / CI flows. `AstroPATAuth(httpx.Auth)` attaches the bearer per request and retries once on 401 with a forced refresh. No writes to `~/.astro/config.yaml`, so we never race the astro CLI's writes
- **Config schema**: `Auth` gains a `kind` discriminator (`basic` / `token` / `astro_pat`) plus optional `context` and `deployment_id` fields. Backward compatible, configs without `kind` infer it from existing fields so existing token / basic instances keep working untouched
- **Adapter wiring**: `AdapterManager.configure(auth_kind=...)` constructs an `AstroPATAuth` instance; the adapter base accepts an optional `auth_handler` (httpx.Auth) that takes precedence over the legacy token_getter / basic_auth_getter path. `detect_version` re-raises `AstroPATError` so missing-login surfaces clearly instead of as a vague version-detection failure
- **Discovery**: dropped `_get_or_create_token`, `create_deployment_token`, the `LEGACY_TOKEN_NAME` warning, and the user-specific token-name machinery. `_deployment_to_instance` records the active astro context on each emitted instance so future `astro context switch` doesn't silently change which credential af uses
- **astro_cli.get_context**: now reads `~/.astro/config.yaml` directly (the canonical source astro itself uses), falling back to parsing `astro context list` only when the file is missing. The old asterisk table parser was unreliable across CLI versions because current astro highlights the active row with ANSI color instead
- **CLI display**: `af instance list` and `af instance current` show `astro pat (<context>)` for PAT-backed instances

## test plan

- [x] `uv run pytest --ignore=tests/integration` (495 / 495 pass, including 31 new tests for the resolver, auth flow, config schema, and adapter-manager wiring)
- [x] `uv run ruff check`, `uv run prek run --all-files` clean
- [x] end-to-end against a real Astro deployment in our Sandbox workspace: `af instance discover astro` produces the right config, `af dags list` and `af api version` succeed via the PAT auth flow, error paths (no astro login, hibernating deployment, ASTRO_HOME pointing at empty dir) surface clear actionable messages
- [x] backward compat: existing token-based instances (`af instance add --token`) keep working unchanged
- [ ] team review of the auth design before un-drafting

## notes on 401 disambiguation

Astro's deployment proxy returns a bare 401 with `Not authorized` for all auth failures (expired token, invalid signature, no workspace access). There's no `WWW-Authenticate` error code we can use to skip the retry when the token is fine but access is missing. v1 always attempts one refresh on 401; if that becomes a problem (broken deployment hammered with requests, Auth0 quota), we add cooldown-after-failed-retry which is more correct than naive debounce. Tracked separately


astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
```diff
@@ -0,0 +1,343 @@
...
+                # API-token flow (or malformed config): surface what's there
+                # with expires_at=0 so callers don't keep retrying.
+                if disk_bearer:
+                    self._cached = _CachedToken(disk_bearer, disk_exp or 0.0)
+                    return disk_bearer
+                raise AstroNotLoggedInError(
+                    f"No refresh_token or token in astro context for {domain!r}. Run `astro login`."
+                )
+
+            new_bearer, new_exp = self._refresh(domain, refresh_token)
```
kaxil, line 268: Auth0 rotates the refresh token by default on every successful refresh-token grant. This code reads `refreshtoken` from disk and exchanges it, but never writes the new refresh token back to `~/.astro/config.yaml`. Next time the astro CLI itself tries to refresh, it'll use the now-invalidated old refresh token and the user's `astro` session breaks (and we keep working until our access token expires, then we break too). At minimum we should detect this (`response["refresh_token"]` differs from what we sent) and either persist it back, or fall through to "tell the user to `astro login`". Right now whichever side refreshes second silently invalidates the other.

astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
```diff
@@ -0,0 +1,343 @@
...
+        self._resolver = resolver
+
+    def auth_flow(self, request):  # type: ignore[no-untyped-def]
+        token = self._resolver.get_token()
+        request.headers["Authorization"] = f"Bearer {token}"
+        response = yield request
+        if response.status_code != 401:
+            return
+        try:
+            new_token = self._resolver.get_token(force_refresh=True)
```
kaxil, line 387: Force-refreshing on every 401 is fine for an expired-token case, but hurts when the user genuinely doesn't have access to a deployment (Astro's deployment proxy returns a bare 401 then too, as the docstring above notes). Each request through this Auth handler will mint a fresh Auth0 token, get 401'd, then mint another one for the next call. For an `af` workflow that walks every deployment in a workspace, that's an Auth0 round-trip per deployment per call. Worth either negative-caching the 401 per-URL for a short window, or only force-refreshing when the cached token is older than some threshold (eg older than the last successful 200 we saw).

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro.py
```diff
@@ -114,42 +90,21 @@ def discover(
         if all_workspaces:
             workspace_map = self._get_workspace_map()
 
+        # Capture the active context once. We only pin to it (record on the
+        # instance) when it differs from DEFAULT_ASTRO_DOMAIN — see the
+        # constant's docstring for the reasoning.
+        active_context = self.get_context()
+        context = active_context if active_context != DEFAULT_ASTRO_DOMAIN else None
```
kaxil: This is a footgun. We discover deployments under `astronomer.io`, persist them with `astro_context=None`, then at request time the resolver follows `astro context` wherever it points. If the user runs `astro context switch` to a dev tenant, our next call to a recorded `astronomer.io` deployment URL will send a dev-tenant bearer to a prod-tenant URL. Best case: 401. Worst case: if the tenants ever share a clientId or the user has access to both, we leak credentials cross-tenant. I'd just always pin `astro_context` to whatever was active at discovery time. The docstring's "matches astro CLI's mental model" argument doesn't outweigh the cross-tenant token leak.

astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
```diff
@@ -0,0 +1,343 @@
...
+                    f"No refresh_token or token in astro context for {domain!r}. Run `astro login`."
+                )
+
+            new_bearer, new_exp = self._refresh(domain, refresh_token)
+            self._cached = _CachedToken(new_bearer, new_exp)
+            return new_bearer
+
+    def _fresh(self, cached: _CachedToken) -> bool:
+        # expires_at=0 means static (eg ASTRO_API_TOKEN); never auto-refresh.
+        if cached.expires_at == 0.0:
```
kaxil: `expires_at == 0.0` is the sentinel for `ASTRO_API_TOKEN` (correct), but also fires for the line 231 path where the on-disk config is malformed and we couldn't parse `expiresin`. Once that fallback caches with `expires_at=0`, `_fresh()` returns True forever and we'll keep sending a possibly-expired bearer until the process restarts. Should distinguish "static API token, never expires" from "I don't know when this expires, treat it as stale". Eg a separate sentinel like `expires_at=-1` for the static case, or a `static: bool` field on `_CachedToken`.

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py
```diff
@@ -301,22 +233,28 @@ def is_installed(self) -> bool:
...
-        Returns:
-            Context domain (e.g., 'cloud.astronomer.io') or None
+        Reads ``~/.astro/config.yaml`` directly (canonical source). Falls
+        back to parsing ``astro context list`` (legacy asterisk format)
+        only when the file is missing or unparseable.
         """
+        # astro CLI's `context list` colors the active row with ANSI codes
+        # rather than marking it with an asterisk, so the table parser is
+        # unreliable across CLI versions.
+        from astro_airflow_mcp.astro_pat import _config_path, _read_yaml
```
kaxil: Importing `_config_path` and `_read_yaml` (both private by convention) from `astro_pat` inline inside `get_context` couples discovery to the internals of the auth module. If `astro_pat` later grows a per-instance config root, or moves yaml parsing out, this breaks silently with no type or import-time signal. Either promote these to a public `astro_pat.read_config()` / `astro_pat.config_path()` (and put them at the top of the file), or move the yaml-reading into a shared `_astro_config.py` module that both consume.

astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
```diff
@@ -0,0 +1,343 @@
...
+        return f"https://{head}.api.{rest}/{_AUTH_CONFIG_PATH}"
+    return f"https://api.{domain}/{_AUTH_CONFIG_PATH}"
+
+
+def _parse_expiry(ctx: dict[str, Any]) -> float:
+    """Return token expiry as epoch seconds. 0 if unknown."""
+    expiresin = ctx.get("expiresin")
+    if isinstance(expiresin, str):
+        try:
+            return datetime.fromisoformat(expiresin).timestamp()
```
kaxil: `datetime.fromisoformat(expiresin).timestamp()` treats the parsed datetime as local time when the string has no tz suffix. Astro CLI writes these without a `Z` (eg `2025-05-01T15:30:00`), so on a machine in `America/Los_Angeles` we'll compute an expiry 7-8 hours off from the actual UTC expiry, meaning we either refresh way too early or, worse, treat an expired token as still fresh. Same issue on line 144 when PyYAML returns a naive datetime. Fix is to assume UTC when naive: `dt.replace(tzinfo=timezone.utc).timestamp()` if `dt.tzinfo is None`.

astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
jlaneve: good catch, fixed in 4bb6d53. on rotation we now grab the same flock astro CLI uses (10s timeout, 100ms poll) and atomic-write the new bearer/refreshtoken/expiresin back. when auth0 doesn't rotate (or returns the same token), we skip the disk write entirely so the hot path stays cheap

astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
jlaneve: fixed in 7570414. when the parsed datetime is naive (which is what astro CLI writes today, no Z, no offset), we now attach UTC before calling .timestamp(). new tests run under TZ=America/Los_Angeles to confirm both the str and PyYAML-datetime branches anchor to UTC instead of local

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro.py
jlaneve: fixed in 3dade7e. you're right, correctness wins. always pinning astro_context now, dropped the DEFAULT_ASTRO_DOMAIN constant. test_discover_returns_pat_instances asserts astronomer.io is recorded; the get_context==None path keeps the None case

astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
jlaneve: fixed in 6ce8950. added a static bool to _CachedToken so the two cases are explicit. ASTRO_API_TOKEN sets static=True (always fresh). the malformed-config path stays static=False with expires_at=disk_exp or 0.0, so a parseable expiry caches normally and a malformed one re-resolves from disk on every call instead of holding stale bytes

astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
jlaneve: fixed in d7b016d. resolver now tracks _last_refresh_at and skips force_refresh if we already refreshed within REFRESH_DEBOUNCE_SECONDS (60s). unit test confirms two quick force_refresh calls produce one /oauth/token request and the second 401 propagates with the cached bearer

astro-airflow-mcp/src/astro_airflow_mcp/discovery/astro_cli.py
jlaneve: fixed in 8bddd9c. extracted the shared helpers (_astro_home, _config_path, _read_yaml, _find_context, _context_key, _parse_expiry, _bearer_from_ctx, _auth_config_url) plus AstroPATError + AstroNotLoggedInError into a new _astro_session module. astro_pat re-exports them for backward compat, discovery now imports from the stable shared location

astro-airflow-mcp/src/astro_airflow_mcp/astro_pat.py
```diff
@@ -0,0 +1,395 @@
...
+                if isinstance(v, dict) and v.get("domain") == domain:
+                    key = k
+                    ctx = v
+                    break
+        if not isinstance(ctx, dict):
+            return
+        ctx["token"] = f"Bearer {bearer}"
+        ctx["refreshtoken"] = refresh_token
+        ctx["expiresin"] = (
+            datetime.fromtimestamp(expires_at, tz=timezone.utc).replace(tzinfo=None).isoformat()
```
kaxil: The fixes for everything I flagged earlier look solid. Verified each one against the new code.

One small follow-up though: this writes `expiresin` as a naive UTC string, but `astro` itself writes it tz-aware with the local offset (eg `2026-05-01T16:48:36.53461+01:00`, I just checked my own `~/.astro/config.yaml`). Functionally fine in both directions (viper reads naive as UTC, PyYAML handles tz-aware), so no bug, but our rotation will silently flip the format in the user's config file from the astro-style to a quoted naive string.

Two related notes while you're in there:
- The `_parse_expiry` docstring in `_astro_session.py` says astro writes "no Z suffix, no offset". That's the assumption the naive-to-UTC fallback is built around, but viper actually serializes with the offset (verified by running a viper write against a config). The naive fallback is still useful as defense-in-depth (eg if a user hand-edits) but the docstring premise is off.
- If you want format parity, passing the tz-aware datetime object directly (`ctx["expiresin"] = datetime.fromtimestamp(expires_at, tz=timezone.utc)`) yields `2026-05-02 16:42:12+00:00` via `safe_dump`, closer to viper's style. Optional polish.

jlaneve: fixed in 2971bf8. _persist_rotated_session now passes a tz-aware datetime so safe_dump emits the offset (eg 2026-05-02 16:42:12+00:00) instead of flipping the format to a quoted naive string. also corrected the _parse_expiry docstring, the naive-to-UTC fallback is defense-in-depth for hand-edits, not the common case (viper writes with offset)

kaxil (approved):

PR #205, opened 2026-05-01
af: catch OSError in _read_yaml so non-file ASTRO_HOME degrades gracefully

jlaneve:
## summary

found while smoke-testing the 0.7.0a1 alpha (#204): `_read_yaml` only caught `FileNotFoundError`, so any other read failure leaked through as an uncaught exception that `detect_version` then masked as "Failed to detect Airflow version"

simplest repro: `ASTRO_HOME=/dev/null af api version` (`/dev/null/config.yaml` raises `NotADirectoryError`, which is an `OSError` but not `FileNotFoundError`). same for permission-denied or broken symlinks

## what changed

- `_read_yaml` now catches `OSError` instead of `FileNotFoundError` (both the initial read and the partial-write retry)
- regression test: `ASTRO_HOME=/dev/null` now surfaces `AstroNotLoggedInError` cleanly

## test plan

- [x] uv run pytest --ignore=tests/integration (490/490)
- [x] uv run prek run --all-files
- [x] manual smoke: `ASTRO_HOME=/dev/null af api version` now reports "No astro context. Run astro login first." instead of "Failed to detect Airflow version"

### DIFF
--- a/astro-airflow-mcp/src/astro_airflow_mcp/_astro_session.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/_astro_session.py
@@ -47,15 +47,18 @@ def _config_path() -> Path:
 
 
 def _read_yaml(path: Path) -> dict[str, Any]:
-    """Read and parse the astro config. Empty/missing/unreadable → {}.
+    """Read and parse the astro config. Empty/missing/non-file → {}.
 
-    Returns an empty dict on any OSError (missing file, permission denied,
-    ASTRO_HOME pointing at a non-directory like ``/dev/null``, etc) so the
-    caller falls through to "no astro session" rather than the raw OS error.
+    Treats "missing or not-a-regular-file" as "no astro session" (path is
+    absent, ASTRO_HOME points at /dev/null or a directory, etc). Other
+    OSError shapes (PermissionError, IOError) are deliberately not
+    swallowed — those mean a real config exists but isn't readable, and
+    surfacing the error is more useful than a misleading "run astro
+    login" downstream.
     """
     try:
         text = path.read_text()
-    except OSError:
+    except (FileNotFoundError, NotADirectoryError, IsADirectoryError):
         return {}
     try:
         return yaml.safe_load(text) or {}
@@ -67,7 +70,7 @@ def _read_yaml(path: Path) -> dict[str, Any]:
         time.sleep(0.05)
         try:
             return yaml.safe_load(path.read_text()) or {}
-        except (OSError, yaml.YAMLError):
+        except (FileNotFoundError, NotADirectoryError, IsADirectoryError, yaml.YAMLError):
             return {}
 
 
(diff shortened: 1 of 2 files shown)


astro-airflow-mcp/src/astro_airflow_mcp/_astro_session.py
```diff
@@ -47,10 +47,15 @@ def _config_path() -> Path:
...
+    """Read and parse the astro config. Empty/missing/unreadable → {}.
+
+    Returns an empty dict on any OSError (missing file, permission denied,
+    ASTRO_HOME pointing at a non-directory like ``/dev/null``, etc) so the
+    caller falls through to "no astro session" rather than the raw OS error.
+    """
     try:
         text = path.read_text()
-    except FileNotFoundError:
+    except OSError:
```
kaxil: Catching all of `OSError` is broader than the `/dev/null` repro needs. `PermissionError` on a real config (e.g. user's `~/.astro/config.yaml` chmod'd 000) will now silently fall through to "Run `astro login` first", which is misleading — the file is there, it's just unreadable. Was the broad catch intentional, or would `(FileNotFoundError, NotADirectoryError, IsADirectoryError)` be enough to fix the reported bug while still surfacing genuine read failures?

astro-airflow-mcp/tests/test_astro_pat.py
```diff
@@ -577,3 +577,13 @@ def test_unparseable_yaml_treated_as_no_session(self, astro_home: Path):
         resolver = AstroPATResolver(env={})
         with pytest.raises(AstroNotLoggedInError):
             resolver.get_token()
+
+    def test_astro_home_at_dev_null_treated_as_no_session(self, monkeypatch):
+        # ASTRO_HOME=/dev/null is a sentinel some wrappers use to "neutralize"
+        # global astro state. Reading /dev/null/config.yaml raises
+        # NotADirectoryError (an OSError, NOT FileNotFoundError); the
+        # resolver should still surface "no session" cleanly.
+        monkeypatch.setenv("ASTRO_HOME", "/dev/null")
```
kaxil: `/dev/null` is Unix-only. `test_config.py::test_load_with_dev_null_via_env` (line 549) already has the right pattern for this: `os.devnull` plus a `pytest.skip` if it's a regular file on the platform. Worth mirroring that here so the test doesn't fail on Windows.

akshayarora (approved):
Yeah, this is a clean and minimal fix with a good explanatory docstring. Thanks!

kaxil (approved):

astro-airflow-mcp/src/astro_airflow_mcp/_astro_session.py
jlaneve: good catch, narrowed in c860abf. now only catches the three OSError shapes that actually mean 'no readable config file' (FileNotFoundError, NotADirectoryError, IsADirectoryError). PermissionError and other OSErrors propagate so chmod 000 surfaces clearly instead of dead-ending at 'run astro login'

astro-airflow-mcp/tests/test_astro_pat.py
jlaneve: fixed in c860abf, mirrors the test_load_with_dev_null_via_env pattern (os.devnull + skip when it's a regular file). thanks for the pointer

PR #214, opened 2026-05-08
update blueprint skill with findings from Otto vs Claude experiment

tayloramurphy:
## Summary

- Grounds the `blueprint` skill in concrete failure modes observed during a recent Otto vs Claude Blueprint experiment ([Slack thread](https://astronomer.slack.com/archives/C0AQGSZ7GGN/p1777935735142399), [doc](https://docs.google.com/document/d/128stMvFPgHxKMhIlq5Ifs8M07pA1aSaARkUtm7AeGz4/edit), full Otto session transcript)
- Adds operating rules and a "Designing the Blueprint Library" section so the agent picks an architecture before writing code, instead of producing a single pipeline-runner blueprint by default
- Documents real bugs the experiment exposed: safe-mode loader discovery, non-YAML-compatible Pydantic types, ``__file__``-relative path breakage on Astro Cloud, schema generation via shell redirect producing invalid JSON (the actual root cause of empty Astro IDE template libraries), missing provider packages in the standalone ``uvx`` CLI environment, Airflow 3 import paths, and customer cluster policy tag requirements
- Generalizes the language so the guidance applies to any orchestration migration and any kind of project-relative artifact, not just Jenkins/SQL

### DIFF
--- a/skills/blueprint/SKILL.md
+++ b/skills/blueprint/SKILL.md
@@ -11,12 +11,24 @@ You are helping a user work with Blueprint, a system for composing Airflow DAGs 
 > **Repo**: https://github.com/astronomer/blueprint
 > **Requires**: Python 3.10+, Airflow 2.5+, Blueprint 0.2.0+
 
+## Operating Rules
+
+These behaviors apply throughout the session:
+
+- **Apply fixes, don't just describe them.** When you identify a problem, edit the file. Don't explain a fix and stop.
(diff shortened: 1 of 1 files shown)


jeremybeard:
Going to close this as I've just updated the Blueprint skill for 0.3.0, which fixed some of the issues noted in here, and at the same time brought in some of the still-relevant info. I've left out the parts that are really about Astro and not specifically about Blueprint.

PR #218, opened 2026-05-28
debugging-dags: Add package version change checks to Step 3

kaxil:
## Summary

Step 3 (Check Context) only mentioned code deploys as a recent-change signal, which misses a really common failure mode: the user's code didn't change but a package they depend on did. Two flavours of that — the worker image was rebuilt and pulled in a new transitive, or a venv-style operator resolved its `requirements` against a moving index at task runtime.

This PR adds instructions for catching both.

## What changed

A new "Package version changes" subsection under Step 3, with three checks the agent should walk through in order.

The first is a worker image diff. When image history exists (every Astro deploy is a new image), `docker run --rm <image> pip freeze` on current vs previous and a `diff` between them is ground truth — anything that changed shows up directly. This is the cleanest signal and the agent should reach for it first.

Image diff is blind to `@task.virtualenv`, `PythonVirtualenvOperator`, `ExternalPythonOperator`, and `KubernetesPodOperator` though, because those build their env per task run, outside the worker image. So the second check is: if the failed task is one of those operators, read its `requirements` (or `image`) arg directly. An unbounded specifier like `pandas>=2.0.0` or an `image="foo:latest"` is the prime suspect, and the fix is to pin.

Third check is an index lookup, for cases where image diff isn't conclusive — usually because there's no image history or it's a venv-style operator. The agent has to detect the configured index first (it may not be PyPI — customers run Artifactory, devpi, etc.) by checking `UV_INDEX_URL` / `PIP_INDEX_URL` / `pyproject.toml` / `pip.conf` / Dockerfile flags, then query that index for releases of the suspect package in the failure window. PyPI's `/pypi/<pkg>/json` shape is widely supported by private indexes, with the Simple API as a fallback.

I also added one bullet to Step 4's Prevention list about pinning deps so this doesn't happen silently in the first place.

## Why this shape

I went back and forth on whether to assume PyPI. Most users *are* on PyPI but enough customers aren't that hardcoding it would produce confidently wrong answers in the cases where it matters most. So the skill asks the agent to detect the index, with the Simple API and "ask the user" as fallbacks.

Kept this PR purely reactive (debugging-time). The obvious follow-up is proactive linting at authoring time — Otto noticing `pandas>=2.0.0` and suggesting `<3.0.0` before it ever ships — but that belongs in `authoring-dags`, not here.

## Credit

Surfaced by Elad Kalif

### DIFF
--- a/skills/debugging-dags/SKILL.md
+++ b/skills/debugging-dags/SKILL.md
@@ -42,12 +42,45 @@ Once you have identified a failed task:
 Gather additional context to understand WHY this happened:
 
 1. **Recent changes**: Was there a code deploy? Check git history if available
-2. **Data volume**: Did data volume spike? Run a quick count on source tables
-3. **Upstream health**: Did upstream tasks succeed but produce unexpected data?
-4. **Historical pattern**: Is this a recurring failure? Check if same task failed before
-5. **Timing**: Did this fail at an unusual time? (resource contention, maintenance windows)
+2. **Package version changes**: Was a package upgraded — in the image, in a venv-style operator, or at the index? See [Package version changes](#package-version-changes) below.
+3. **Data volume**: Did data volume spike? Run a quick count on source tables
+4. **Upstream health**: Did upstream tasks succeed but produce unexpected data?
+5. **Historical pattern**: Is this a recurring failure? Check if same task failed before
+6. **Timing**: Did this fail at an unusual time? (resource contention, maintenance windows)
 
 Use `af runs get <dag_id> <dag_run_id>` to compare the failed run against recent successful runs.
+
+### Package version changes
+
+A common cause of failures with no git activity is dependency drift — the user's code didn't change, but a package they depend on did. Check in this order:
+
+1. **Worker image diff** (preferred when available). Every Astro deploy = new image tag, so the registry has a "before" and "after". Diff `pip freeze` between current and previous image — that's ground truth for what changed:
+   ```
+   docker run --rm <current_image> pip freeze > /tmp/now.txt
+   docker run --rm <previous_image> pip freeze > /tmp/prev.txt
+   diff /tmp/prev.txt /tmp/now.txt
+   ```
+   `af config providers` lists currently installed provider versions — useful for cross-checking against modules named in the traceback.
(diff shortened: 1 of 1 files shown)


skills/debugging-dags/SKILL.md
```diff
@@ -42,13 +42,46 @@ Once you have identified a failed task:
...
+   - `~/.pip/pip.conf`, `/etc/pip.conf`
+   - `Dockerfile` `--index-url` flags
+
+   Then query for releases of the suspect package since the first failure started. PyPI:
+   ```
+   curl -s https://pypi.org/pypi/<pkg>/json | jq '.releases | to_entries | map({version: .key, uploaded: .value[0].upload_time}) | sort_by(.uploaded) | reverse | .[:5]'
+   ```
+   Private indexes usually expose the same `/pypi/<pkg>/json` shape; fall back to the Simple API (`/simple/<pkg>/`) or ask the user if neither works.
+
+A release timestamp landing between the last green run and the first red run, for a package named in the traceback, is the answer.
```
eladkal, line 84: this is great!
lets add also a case of change in Python versions. Sometimes a recent change to Python version is the root cause behind library version changes. That can be local (in a specific venv operator) or a change in Python version in the whole enviroment.

eladkal (approved):

kaxil:
CI failure is fixed by https://github.com/astronomer/agents/pull/219

skills/debugging-dags/SKILL.md
kaxil: good catch, wasn't covered. pushed 40ed1aa.

for the image diff bullet I added `docker run --rm <image> python --version` between current and previous, since `pip freeze` doesn't surface the interpreter itself and a minor bump can break otherwise-identical wheels.

on venv-style operators I extended the args to read to include `python_version` (`@task.virtualenv` / `PythonVirtualenvOperator`) and `python` (`ExternalPythonOperator`), plus a fix bullet about pinning fully-qualified versions like `"3.11.7"` over `"3.11"` so this can't drift silently.

akshayarora (approved):
This looks great

PR #221, opened 2026-06-09
feat: Return structured tool errors and add MCP safety annotations

kaxil:
## Summary

MCP tools previously collapsed every failure to a bare `str(e)`. An agent calling, say, `get_task_instance("my_dag", "run_1", "extract")` got back `"404 Not Found"` with no way to tell a **wrong `dag_id`** (fix the argument and retry) from **Airflow being unreachable** (stop, surface to the user) — the same opaque string, opposite correct responses.

This PR makes every tool return a **structured error payload** and stamps **MCP safety annotations** on all 34 tools.

### Structured errors-as-data

A new `tool_errors` module turns exceptions into:

```json
{
  "error": "Endpoint not found: dags/my_dag",
  "error_type": "NotFoundError",
  "hint": "The resource was not found. Check dag_id/dag_run_id/task_id — a list_* tool can confirm valid values.",
  "retryable": true,
  "dag_id": "my_dag"
}
```

- `retryable` distinguishes caller-fixable problems (bad arg, 400/404/429) from hard blocks (auth, 5xx, read-only, unreachable), so the agent knows whether a corrected retry can help.
- `hint` gives the next action; identifying arguments are echoed back so the model has what it needs to self-correct.
- Classifies `ReadOnlyError`, `NotFoundError`, Astro auth errors, `httpx` status codes (400/401/404/409/429/5xx), timeouts, and transport errors.
- `str(e)` is always preserved inside `error`, so anything that scanned the old bare-string responses keeps working.

The 30 `except Exception: return str(e)` sites, the consolidated tools' per-section error embeds, and `trigger_dag_and_wait`'s trigger-failure path all use this now.

### MCP safety annotations

`read_only()` / `write(...)` helpers (default-cautious) stamp `ToolAnnotations` on every tool: 27 reads as read-only + idempotent; 7 writes with accurate hints — `delete_dag_run` destructive + non-idempotent, `clear_*` destructive + idempotent, `pause`/`unpause` non-destructive + idempotent, `trigger*` non-destructive + non-idempotent. Clients can gate writes **before** invocation, complementing the existing `AF_READ_ONLY` guard (which only refuses *after* the model has decided to call).

## Design rationale

- **Why keep `str(e)` in `error`?** Backward compatibility — existing tests and any consumer that grepped the old string still find their substrings. New consumers get the structured fields.
- **Why a `retryable` boolean instead of leaving it to the model?** The retry/stop decision is the single highest-value signal for agent self-correction and shouldn't depend on the model re-deriving it from prose each time.
- **Why default-cautious annotations?** A tool is only advertised "safe" when it explicitly opts in via `read_only()`; a new write tool that forgets to annotate is treated as destructive, not auto-approvable.
- **`normalize_airflow_url` now strips `user:password@`** from the netloc (preserving host/port/IPv6 verbatim). Credentials are supplied separately via the auth getters, and the normalized URL can appear in logs and in the URL embedded in error messages now surfaced to the model.

## Usage example

```python
# Read tools auto-approvable; writes gated
result = await client.call_tool("get_dag_details", {"dag_id": "nope"})
# -> {"error": "...", "error_type": "NotFoundError", "hint": "...", "retryable": true, "dag_id": "nope"}
```

## Tradeoffs / known issues

- When Airflow is unreachable, the adapter wraps the `httpx.ConnectError` in a bare `RuntimeError` ("Ensure Airflow is running and accessible") with no chained cause, so the classifier falls to the generic hint and `retryable=false`. That is correct behavior, and the message itself carries the guidance; a follow-up could chain the cause in `detect_version` and classify on it for a sharper hint.
- The `"No DAGs found. Response: …"`-style fallbacks on the *success* path (API returned an unexpected shape) remain plain strings; they are not error paths and are out of scope here.

### DIFF
--- a/astro-airflow-mcp/src/astro_airflow_mcp/tool_annotations.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/tool_annotations.py
@@ -1,0 +1,62 @@
+"""MCP tool safety annotations (read-only / destructive hints).
+
+These map onto the MCP ``ToolAnnotations`` spec so a client can decide, *before*
+invoking a tool, whether it is safe to auto-approve (reads) or must be gated
+behind user confirmation (writes / destructive operations). This complements the
+server-side ``AF_READ_ONLY`` guard, which only refuses a write *after* the model
+has already decided to call it.
+
+Defaults are cautious: a tool is only advertised as safe when it explicitly opts
+in via :func:`read_only`. Writes default to destructive + non-idempotent and are
+narrowed by the caller when a weaker hint is accurate (e.g. ``pause_dag`` is a
+non-destructive, idempotent write).
+
+``openWorldHint`` is True for every tool here: they all talk to an external
+Airflow instance whose state is outside this process.
+"""
+
+from __future__ import annotations
+
+from mcp.types import ToolAnnotations
+
+
+def read_only(title: str | None = None) -> ToolAnnotations:
+    """Annotations for a tool that only reads Airflow state.
+
+    Safe for clients to auto-approve. Reads are idempotent by definition.
+    """
+    return ToolAnnotations(
+        title=title,
+        readOnlyHint=True,
+        destructiveHint=False,
+        idempotentHint=True,
+        openWorldHint=True,
+    )
+
+
+def write(
+    title: str | None = None,
+    *,
+    destructive: bool = True,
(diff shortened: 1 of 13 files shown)


schnie (approved):
Nice enhancements

PR #229, opened 2026-06-16
fix(diagnostic): page through all task instances in diagnose_dag_run

### LINKED ISSUE #228: diagnose_dag_run misses failed task instances beyond the first 100 (no pagination / no state filter)
## Summary

`diagnose_dag_run` (and the `_get_failed_task_instances` helper behind `clear_dag_run`/run tooling) can report **zero failed tasks for a run that has failed tasks**, whenever the failing task instances fall beyond the first 100 instances of the DAG run. This is silent — no truncation marker — so an agent consuming the result concludes the run is healthy.

This is especially likely on DAGs with **dynamically mapped tasks**, where a single run easily exceeds 100 task instances and a failure can sit at a high `map_index`.

## Affected versions

Present in `0.6.4` and **still present on `main` (the latest `0.8.2` line)**. Bumping the version does not fix it.

## Root cause

`diagnose_dag_run` fetches task instances in a single page and filters in memory:

```python
# tools/diagnostic.py
tasks_data = adapter.get_task_instances(dag_id, dag_run_id)   # no limit/offset/state
task_instances = tasks_data.get("task_instances", [])
for ti in task_instances:
    if state in ("failed", "upstream_failed"):
        failed_tasks.append({...})
```

The adapter method defaults to a single request of 100 rows, offset 0, with no `state` filter (identical in `adapters/airflow_v2.py` and `adapters/airflow_v3.py`):

```python
def get_task_instances(self, dag_id, dag_run_id, limit: int = 100, offset: int = 0):
    return self._call(f"dags/{dag_id}/dagRuns/{dag_run_id}/taskInstances",
                      params={"limit": limit, "offset": offset})
```

`limit`/`offset` exist but **no caller ever varies them and there is no pagination loop**. The two internal callers (`diagnostic.py::diagnose_dag_run` and `dag_run.py::_get_failed_task_instances`) both call with the defaults. There is also no MCP tool that lists task instances with `limit`/`offset` exposed, so a client cannot page through them either.

## Reproduction

1. Take a DAG run with > 100 task instances (e.g. a couple of dynamically mapped tasks).
2. Make a task fail at a `map_index` that places it past the 100th instance in the listing order.
3. Call `diagnose_dag_run`.
4. **Observed:** `state_counts` sums to 100, `failed_tasks: []`, run looks healthy.
   **Expected:** the failed mapped instance appears in `failed_tasks`.

## Suggested fix

In `diagnose_dag_run` / `_get_failed_task_instances`, either:

- **(a)** paginate: loop `offset` by `limit` until `total_entries` is exhausted; or
- **(b)** push the filter server-side: pass `state=["failed", "upstream_failed"]` to the `taskInstances` call so only failures are returned (small payload regardless of run size). This requires adding a `state` kwarg to the adapter `get_task_instances`.

Option (b) is cheapest and most robust for the diagnose path.

## Related secondary gap

`get_task_instance` (singular) hits `…/taskInstances/{task_id}` with no `map_index`, so it 404s on a mapped instance — there is no way to fetch a single mapped instance's *state* (only `get_task_logs` accepts `map_index`). Adding `map_index` to the singular getter would round this out.

## Code references

- `astro-airflow-mcp/src/astro_airflow_mcp/tools/diagnostic.py` — `diagnose_dag_run`
- `astro-airflow-mcp/src/astro_airflow_mcp/tools/dag_run.py` — `_get_failed_task_instances`
- `astro-airflow-mcp/src/astro_airflow_mcp/adapters/airflow_v3.py` / `airflow_v2.py` — `get_task_instances`

gallejesus:
## Summary

Fixes #228.

`diagnose_dag_run` (and the `_get_failed_task_instances` helper) read a **single page** of task instances — `get_task_instances(dag_id, dag_run_id)` defaults to `limit=100, offset=0` with no pagination. On a DAG run with many **dynamically mapped** task instances, a failed instance at a high `map_index` falls past the first 100, so it never enters `failed_tasks` / `state_counts` and the run is reported as healthy with zero failures.

## Changes

- **`adapters/base.py`** — add a concrete `get_all_task_instances()` that pages through the existing `get_task_instances()` primitive until a short page is returned, merging the results. It's a composition over an existing endpoint (not a new one), so it lives as one concrete method on the base class and works unchanged for both Airflow 2 and 3 — no per-version duplication.
- **`tools/diagnostic.py`** and **`tools/dag_run.py`** — both callers now use `get_all_task_instances()`.
- **`tools/diagnostic.py`** — include `map_index` in the `failed_tasks` payload. This does not affect which failures are found (the pagination does that); it lets a consumer tell apart multiple failed mapped instances that share the same `task_id`.

## Tests

- `tests/test_adapters.py::TestGetAllTaskInstances` — paging merges across pages (incl. a failure on a later page), a short first page stops immediately, and a payload with no `task_instances` yields an empty result.
- `tests/test_consolidated_tools.py` — regression test asserting a failed mapped instance beyond the first page surfaces in `failed_tasks`.

Run with `uv run pytest tests/ --ignore=tests/integration/`.

### DIFF
--- a/astro-airflow-mcp/src/astro_airflow_mcp/adapters/base.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/adapters/base.py
@@ -430,6 +430,28 @@ class AirflowAdapter(ABC):
             offset: Offset for pagination
         """
 
+    def get_all_task_instances(
+        self, dag_id: str, dag_run_id: str, page_size: int = 100
+    ) -> dict[str, Any]:
+        """Return every task instance for a DAG run, paging past the limit.
+
+        ``get_task_instances`` returns a single page (``page_size`` 100 by
+        default). On a run with many dynamically mapped instances, a failure at
+        a high ``map_index`` falls past the first page and is silently missed.
+        This pages through ``get_task_instances`` until a short page is
+        returned, so it works unchanged for every adapter version.
+        """
+        merged: list[dict[str, Any]] = []
+        offset = 0
+        while True:
+            page = self.get_task_instances(dag_id, dag_run_id, limit=page_size, offset=offset)
+            batch = page.get("task_instances", [])
+            merged.extend(batch)
+            offset += len(batch)
+            if len(batch) < page_size:
+                break
+        return {"task_instances": merged, "total_entries": len(merged)}
+
     @abstractmethod
     def get_task_logs(
         self,
--- a/astro-airflow-mcp/src/astro_airflow_mcp/tools/dag_run.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/tools/dag_run.py
@@ -109,7 +109,9 @@ def _get_failed_task_instances(
     """
     try:
         adapter = _get_adapter()
-        data = adapter.get_task_instances(dag_id, dag_run_id)
+        # Page through all task instances so failures past the first page
+        # (e.g. a high map_index on a large mapped DAG run) are not missed.
+        data = adapter.get_all_task_instances(dag_id, dag_run_id)
         task_instances = data.get("task_instances", [])
         return extract_failed_tasks(task_instances)
     except Exception:
--- a/astro-airflow-mcp/src/astro_airflow_mcp/tools/diagnostic.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/tools/diagnostic.py
@@ -205,7 +205,9 @@ def diagnose_dag_run(dag_id: str, dag_run_id: str) -> str:
 
     # Get task instances for this run
     try:
-        tasks_data = adapter.get_task_instances(dag_id, dag_run_id)
+        # Page through every task instance: a single page (100) can hide a
+        # failed mapped instance at a high map_index on large DAG runs.
+        tasks_data = adapter.get_all_task_instances(dag_id, dag_run_id)
         task_instances = tasks_data.get("task_instances", [])
         result["task_instances"] = task_instances
 
@@ -219,6 +221,7 @@ def diagnose_dag_run(dag_id: str, dag_run_id: str) -> str:
                 failed_tasks.append(
                     {
                         "task_id": ti.get("task_id"),
+                        "map_index": ti.get("map_index"),
                         "state": state,
                         "start_date": ti.get("start_date"),
                         "end_date": ti.get("end_date"),
--- a/astro-airflow-mcp/tests/test_adapters.py
+++ b/astro-airflow-mcp/tests/test_adapters.py
@@ -1525,3 +1525,66 @@ class TestAdapterManagerSSL:
         mgr.get_adapter()
 
         assert mock_create.call_args[1]["verify"] is True
+
+
+class TestGetAllTaskInstances:
+    """Tests for AirflowAdapter.get_all_task_instances pagination helper."""
+
+    def _adapter(self):
+        return AirflowV3Adapter("http://localhost:8080", "3.1.0")
+
+    def test_pages_through_all_results(self, mocker):
+        """It walks every page and merges results past the first page.
+
+        Regression for the bug where a failed instance at a high map_index
+        (beyond page 1) was silently dropped because only one page was read.
+        """
+        adapter = self._adapter()
+        page1 = {
+            "task_instances": [
+                {"task_id": "t", "map_index": 0, "state": "success"},
+                {"task_id": "t", "map_index": 1, "state": "success"},
+            ],
+            "total_entries": 3,
+        }
+        page2 = {
+            "task_instances": [
+                {"task_id": "t", "map_index": 14, "state": "failed"},
+            ],
+            "total_entries": 3,
+        }
+        mock_get = mocker.patch.object(adapter, "get_task_instances", side_effect=[page1, page2])
+
+        result = adapter.get_all_task_instances("dag", "run", page_size=2)
+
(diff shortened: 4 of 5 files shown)


astro-airflow-mcp/src/astro_airflow_mcp/tools/diagnostic.py
```diff
@@ -205,7 +205,9 @@ def diagnose_dag_run(dag_id: str, dag_run_id: str) -> str:
 
     # Get task instances for this run
     try:
-        tasks_data = adapter.get_task_instances(dag_id, dag_run_id)
+        # Page through every task instance: a single page (100) can hide a
+        # failed mapped instance at a high map_index on large DAG runs.
+        tasks_data = adapter.get_all_task_instances(dag_id, dag_run_id)
         task_instances = tasks_data.get("task_instances", [])
         result["task_instances"] = task_instances
```
kaxil, line 212: `result["task_instances"]` now holds **every** task instance (previously capped at one page of ~100). On the large dynamically-mapped runs this fix targets, with thousands of instances, `diagnose_dag_run` serializes the full list into the tool's JSON response, which is fed to the model. That can be hundreds of thousands of tokens of mostly-succeeded instances, and may blow the context window in exactly the scenario the fix is for.

The `summary` block already carries the complete counts and failed-task list needed to fix #228, so consider capping what gets emitted here (e.g. failed instances plus first N, plus a `total`/`truncated` indicator) while keeping `summary` complete.

astro-airflow-mcp/src/astro_airflow_mcp/adapters/base.py
```diff
@@ -430,6 +430,28 @@ def get_task_instances(
...
+        returned, so it works unchanged for every adapter version.
+        """
+        merged: list[dict[str, Any]] = []
+        offset = 0
+        while True:
+            page = self.get_task_instances(dag_id, dag_run_id, limit=page_size, offset=offset)
+            batch = page.get("task_instances", [])
+            merged.extend(batch)
+            offset += len(batch)
+            if len(batch) < page_size:
```
kaxil, line 451: The termination check `len(batch) < page_size` assumes the server hands back exactly what you asked for. Airflow clamps `limit` to `[api] maximum_page_limit` (default 100), so on a deployment that sets that config below `page_size` the first page comes back short, the loop breaks after one page, and anything past that boundary is silently dropped, which re-introduces #228 at a lower threshold. And since `total_entries` is recomputed as `len(merged)`, a caller can't tell the result was truncated.

Paging until `len(merged) >= page.get("total_entries", 0)`, with an empty-batch guard so it can't spin forever, sidesteps that. The same loop also ends quietly if a mid-pagination call returns the `{"available": False}` not-found payload, and nothing bounds the page count if a run is enormous.

astro-airflow-mcp/src/astro_airflow_mcp/tools/dag_run.py
```diff
@@ -109,7 +109,9 @@ def _get_failed_task_instances(
     """
     try:
         adapter = _get_adapter()
-        data = adapter.get_task_instances(dag_id, dag_run_id)
+        # Page through all task instances so failures past the first page
+        # (e.g. a high map_index on a large mapped DAG run) are not missed.
+        data = adapter.get_all_task_instances(dag_id, dag_run_id)
         task_instances = data.get("task_instances", [])
         return extract_failed_tasks(task_instances)
```
kaxil, line 116: Pagination is correctly applied here, but the `map_index` disambiguation added to `diagnose_dag_run` doesn't reach this path: `extract_failed_tasks` (`utils.py:67`) doesn't carry `map_index`, so `trigger_dag_and_wait` still collapses multiple failed instances that share a `task_id`. Adding `"map_index": task.get("map_index")` to `extract_failed_tasks` fixes both the MCP and CLI trigger-wait paths at once and keeps the output consistent with `diagnose_dag_run`.

kaxil (commented):
The core fix looks right. `get_all_task_instances` lives once on the base adapter as a composition over the existing `get_task_instances` (so no per-version duplication), both MCP callers are migrated, and the regression tests cover it well.

The main thing I'd flag is that the CLI still has the bug. `cli/runs.py` carries two hand-duplicated copies of this logic, neither touched here, and it's live code: `af runs diagnose` and `af runs trigger-wait` reach it via `app.add_typer(runs_module.app, name="runs")`. At `cli/runs.py:373` the `diagnose` command still calls single-page `adapter.get_task_instances(...)`, so a failed mapped instance past map_index 100 gets dropped and the run reads as healthy. Same #228 symptom, just on the CLI. And `cli/runs.py:281` (`_get_failed_task_instances`, used by `trigger-wait` at line 259) reads a single page too. Both want `get_all_task_instances`.

That duplication is really why the fix only landed in two places: the same diagnose/failed-task logic exists in `tools/diagnostic.py`, `tools/dag_run.py`, `cli/runs.py`, and `utils.extract_failed_tasks`. Folding it into one shared helper would mean pagination (and `map_index`) only has to be added once.

Same story for `map_index`: `cli/runs.py:384` omits it from the CLI `failed_tasks` dict, and `utils.extract_failed_tasks` (`utils.py:67`) omits it too (see the note on `dag_run.py`), so `trigger_dag_and_wait` still can't tell two failed instances of the same task apart.

Otherwise the MCP `diagnose_dag_run` path is correct; I'd just want the two CLI sites and the payload size (inline) sorted before this merges.

PR #234, opened 2026-06-22
Add skills for authoring tasks with Java SDK

uranusjr:
These are for AIP-108, which will be released with Airflow 3.3. We also plan to add skills for Go SDK, so the skills here are structured specifically to avoid duplication across SDKs.


kaxil (commented):
This is in good shape, and notably self-contained: every cross-reference resolves. The four new skills point at each other and at `authoring-dags` / `deploying-airflow` / `setting-up-astro-project`, all of which are already on `main`. That's the right pattern, and worth calling out because the Go side (#239) references `authoring-language-sdk-tasks` and the shared config file that only exist in this PR, so those links dangle on `main` until this merges.

The version hygiene is good too: `${version}` placeholders, the `-SNAPSHOT`/preview note, and the "read the generated Dockerfile, don't pin from memory" hedge all keep it from going stale. For contrast, #239's deploy skill hardcodes `apache/airflow:3.3.0` and a `@v1.0.0` pin that doesn't resolve, so it could borrow this approach.

The one thing to coordinate is the shared file. This PR and #239 both add `skills/configuring-airflow-language-sdks/SKILL.md`, and #239's version is a Go-inclusive superset of this Java-framed one. Since this PR is the foundation the Go skills build on, the cleanest path is to land this first and have #239 rebase its copy into a MODIFY that adds the Go coordinator. Noted inline.

Two minor consistency points: the descriptions open imperative (Write / Configure / Build and deploy) where the post-#237 catalog settled on third-person-present (Writes / Configures / Builds and deploys), and the prose uses `Dag`/`Dags` while the rest of the catalog uses `DAG`. Both are consistency nits rather than rules.

jason810496 (approved):

skills/deploying-java-sdk-bundles/SKILL.md
```diff
@@ -87,11 +87,64 @@ Unlike Gradle, Maven does **not** validate `mainClass` at build time; a wrong va
...
+For task log records to reach Airflow's log store (and the task log view in the UI), the bundle must include **exactly one** SDK logging artifact per logging facade you use. Versions are managed by `airflow-sdk-bom`; Maven users apply the same artifact IDs.
 
-While the SDK is pre-release, the artifacts and the Gradle plugin may resolve only from Apache's snapshot Nexus. Add it in **both** `pluginManagement` (in `settings.gradle`) and project `repositories` (in `build.gradle`):
+| Facade | Artifact | Setup beyond the dependency |
+|--------|----------|-----------------------------|
+| `System.Logger` (JPL) | `airflow-sdk-jpl` | None — the provider is discovered via `ServiceLoader`. |
+| SLF4J 2.x | `airflow-sdk-slf4j` | None — the binding is discovered automatically (pulls in `slf4j-api` for you). |
+| Log4j 2 | `airflow-sdk-log4j2` | `log4j-core` on the runtime classpath + `AirflowAppender` declared in `log4j2.xml` (below). |
+| `java.util.logging` (JUL) | `airflow-sdk-jul` | Call `AirflowJulHandler.setup()` in `main()` (below), or use a `logging.properties` file (see **configuring-airflow-language-sdks**). |
```
uranusjr, line 101: We should probably try to steer people away from old tech. Maybe add a paragraph that recommends JPL for a green field project, but select a specific logging backend if it’s already used in the libraries the user wants to integrate with.

The preference should be

JPL > SLF4J = Log4j2 > JUL.

kaxil: @jason810496 ^^ ? and could you resolve merge conflicts please

PR #241, opened 2026-07-08
fix: Prevent FastMCP 3.4.3 host guard from 421-ing the plugin endpoint

kaxil:
## Summary

FastMCP 3.4.3 added `HostOriginGuardMiddleware`, enabled by default: it only accepts loopback `Host` headers (`127.0.0.1`/`localhost`/`::1`) plus the ASGI `scope["server"]` host, and returns `421 Misdirected Request` for anything else.

In plugin mode the MCP app is embedded in the Airflow webserver and reached over the Deployment's own hostname, so with an unpinned install that resolves fastmcp 3.4.3, every `/mcp/v1/` request is rejected with `421`. (The AF2 Flask bridge also builds the ASGI scope with no `server` key, so the scope-server fallback doesn't apply.)

This keeps the guard enabled but scoped to the Deployment's hostname, restoring `/mcp/v1/` to a normal `200` MCP handshake. Standalone/local mode is untouched.

## What changed

- `plugin.py`: both `mcp.http_app()` calls (AF3 FastAPI + AF2 Flask) keep `HostOriginGuardMiddleware` **enabled**, scoped to the Deployment hostname derived from `AIRFLOW__WEBSERVER__BASE_URL` (an env var, so present at import — Airflow's `conf` isn't yet). Falls back to `host_origin_protection=False` only when no hostname is derivable (non-Astro embeddings).
- `ASTRO_MCP_ALLOWED_HOSTS` (comma-separated) overrides the derived allowlist — custom domains or extra hosts.
- Pin `fastmcp>=3.4.3,<4`: the kwargs only exist from 3.4.3, and the upper bound stops the next default-behaviour change from silently breaking installs.

## Design rationale

- **Scope the guard vs. disable it:** keeping the guard on preserves DNS-rebinding protection. `base_url` is Astro's canonical hostname source — both the deployment ingress route (Host preserved, no authority rewrite) and the auth-proxy's `changeOrigin` converge on that same per-deployment hostname — so scoping to it matches the `Host` that actually arrives, and it auto-tracks any future hostname change rather than going stale.
- **Custom domains / other Host values:** handled by `ASTRO_MCP_ALLOWED_HOSTS`; the disable-fallback covers non-Astro embeddings where no hostname is known.
- **Scope:** only the embedded/plugin path changes. Standalone (`__main__.py`, `mcp.run`) keeps FastMCP's loopback default, since a locally-bound server is what DNS-rebinding protection is for.

## Interim workaround (no plugin upgrade needed)

Pin `fastmcp<3.4.3` in the Astro project's `requirements.txt` and redeploy.

### DIFF
--- a/astro-airflow-mcp/pyproject.toml
+++ b/astro-airflow-mcp/pyproject.toml
@@ -5,7 +5,10 @@ description = "A FastMCP server for Airflow integration that can run standalone 
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "fastmcp>=0.1.0",
+    # Pinned to a bounded range: 3.4.3 introduced HostOriginGuardMiddleware and the
+    # host_origin_protection/allowed_hosts kwargs the plugin relies on (see plugin.py).
+    # The upper bound guards against another default-behavior change in a new major.
+    "fastmcp>=3.4.3,<4",
     "filelock>=3.0.0",
     "httpx>=0.27.0",
     "packaging>=21.0",
--- a/astro-airflow-mcp/src/astro_airflow_mcp/plugin.py
+++ b/astro-airflow-mcp/src/astro_airflow_mcp/plugin.py
@@ -12,6 +12,7 @@ import logging
 import os
 import threading
 from typing import Any
+from urllib.parse import urlparse
 
 from astro_airflow_mcp import __version__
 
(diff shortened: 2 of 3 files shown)


schnie:
Does Astro Cloud inject the hostname into the webserver/apiserver env vars? Wonder if we could keep the option but pass it the real hostname? Seems like something that would be injected but not sure.

kaxil:
> Does Astro Cloud inject the hostname into the webserver/apiserver env vars? Wonder if we could keep the option but pass it the real hostname? Seems like something that would be injected but not sure.

Updated

schnie (approved):

PR #251, opened 2026-08-03
Cover submit+poll via sensor pattern in airflow-state-store skill

amoghrajesh:
Ref: https://lists.apache.org/thread/xhlwpy4mpptw842dfv0dqyxwkzxq3936

Per community discussion on submit+poll patterns, recommend Airflow's built-in durable execution mode instead of a hand-rolled submit task + sensor/poll task split, for cases where a Triggerer isn't deployed but the job can run synchronously. Adds a Triggerer- and job-duration- conditioned decision tree to the airflow-state-store skill, fixes the skill's proactive trigger so it actually fires when reviewing this DAG shape, and clarifies that ResumableJobMixin's crash-safety only applies
to a single-task submit+poll, not an already-split submit/sensor pair and recommends it as needed.

Net result now:
image

### Before my skill update:  

[before.md](https://github.com/user-attachments/files/30662258/before.md)

### After my skill update

[after.md](https://github.com/user-attachments/files/30662268/after.md)

### DIFF
--- a/skills/airflow-state-store/SKILL.md
+++ b/skills/airflow-state-store/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: airflow-state-store
-description: Persists task and asset state across retries and DAG runs using Airflow 3.3's AIP-103 key/value stores (`task_state_store`, `asset_state_store`) and the crash-safe `ResumableJobMixin`. Use when the user asks about task state store, checkpointing in tasks, persisting state across retries, job IDs surviving worker crashes, watermarks, asset metadata, resumable tasks, crash-safe operators, or "what's new in Airflow 3.3". Also use proactively when reading a DAG that uses Variables or XCom for intra-task coordination state — flag the anti-pattern and recommend task_state_store or asset_state_store instead. Requires Airflow 3.3+.
+description: Persists task and asset state across retries and DAG runs using Airflow 3.3's AIP-103 key/value stores (`task_state_store`, `asset_state_store`) and the crash-safe `ResumableJobMixin`. Use when the user asks about task state store, checkpointing in tasks, persisting state across retries, job IDs surviving worker crashes, watermarks, asset metadata, resumable tasks, crash-safe operators, or "what's new in Airflow 3.3". Also use proactively when reading a DAG that uses Variables or XCom for intra-task coordination state — flag the anti-pattern and recommend task_state_store or asset_state_store instead. Also use proactively when reviewing ANY DAG that submits a job to an external system and waits for it to finish — Databricks, Snowflake, BigQuery, Redshift, Spark, dbt Cloud, EMR, AWS Batch, etc. — whether that is one submit-and-wait operator or split across a separate submit task plus a sensor/polling task; this covers `wait_for_termination`, `deferrable`, `durable`, hand-rolled sensors polling a run/job id, and whether to collapse a submit+sensor split into one task. Requires Airflow 3.3+.
 ---
 
 # Airflow Task State Store (AIP-103)
@@ -42,8 +42,11 @@ When the user asks to review a DAG or asks "is there a better way", scan for the
 | `context["ti"].xcom_push(key="job_id", ...)` to survive retries | XCom is scoped to a DAG run, not a retry; a new ti_id is issued per retry | `task_state_store` or `ResumableJobMixin` |
 | Manual `if Variable.get("job_id"): reconnect else: submit` retry-resume logic | Reimplements what `ResumableJobMixin` already provides, without the crash-safety guarantee | `ResumableJobMixin` |
 | `Variable.set("last_processed_at", ...)` for watermarks | Global; any DAG or task can overwrite it; no scoping to asset | `asset_state_store` |
+| Separate `submit` task (`wait_for_termination=False` / fire-and-forget) + a second sensor/polling task waiting on the same external job (Databricks, Snowflake, BigQuery, Redshift, Spark, etc.) | A green submit task only means the job was *accepted*, not that it *succeeded* — only the sensor task's outcome reflects reality. Whether the split is also wasteful depends on Triggerer availability — see below. | **Triggerer deployed:** collapse to **one task** with `deferrable=True` (implies `wait_for_termination=True`). **No Triggerer, job OK to run synchronously:** collapse to **one task** using the operator's `durable=True` mode instead. **No Triggerer, job genuinely long/worker slots scarce:** see below — keep the split. |
 
 Show a before/after snippet when flagging. Use the canonical examples in Steps 3–5 as the "after".
+
+**The submit+sensor split is worth a comment even when the sensor code is bug-free** — reviewing the sensor's code quality (correct `mode=`, correct terminal-state handling, cached hook) is a separate question from whether the two-task split is the right architecture. The full decision tree (Triggerer deployed / no Triggerer + sync-OK / no Triggerer + long-running) lives in one place — **Section 6, "Submit-and-poll DAGs: one task or two?"** — do not re-derive it here; follow that section.
 
 ---
 
@@ -204,7 +207,9 @@ def load(asset_state_store=None):
 
 ## Section 5 — `ResumableJobMixin`: crash-safe external job submission
 
-Use when an operator submits a job to an external system (Spark, Databricks, dbt Cloud, AWS Batch, etc.) and then polls for completion. Without this mixin, a worker crash during polling means the next retry submits a duplicate job.
+Use when **a single operator/task** submits a job to an external system (Spark, Databricks, dbt Cloud, AWS Batch, etc.) and then polls for completion *within one `execute()` call*. Without this mixin, a worker crash during polling means the next retry of that same task re-enters `execute()` from the top and submits a duplicate job.
+
+**Scope check before recommending it:** if submit and poll are already two *separate* tasks (e.g. a `submit` task handing a job id to a downstream sensor/poll task via XCom), this failure mode doesn't exist in the first place — don't recommend the mixin just because "submit + poll" appears somewhere in the DAG. See the table row below for why.
 
 **When NOT to use `ResumableJobMixin`:**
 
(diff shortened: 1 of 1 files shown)


skills/airflow-state-store/SKILL.md
```diff
@@ -302,9 +308,49 @@ class MySparkOperator(BaseOperator, ResumableJobMixin):
...
+
+A common DAG shape for Databricks/Snowflake/BigQuery/Redshift/Spark jobs: a `submit` task that fires the job with `wait_for_termination=False` (or equivalent) and returns immediately, followed by a hand-rolled sensor task that polls the same job to completion. The submit task going green the instant the external system *accepts* the job (not when it finishes) is always worth pointing out — a retry, alert, or SLA on the submit task alone tells you nothing about real job success, only the sensor task's outcome does. But whether the split itself should be removed depends on Triggerer availability:
+
+**If a Triggerer is deployed** — collapse to one task with `deferrable=True`. This frees the worker during polling *and* the task's own outcome reflects the real job result. Strictly better than the split; no tradeoff to weigh.
+
+```python
+run_job = DatabricksSubmitRunOperator(
+    task_id="run_job",
+    tasks=[{"task_key": "job", "notebook_task": {"notebook_path": NOTEBOOK_PATH}}],
+    deferrable=True,   # wait_for_termination=True is implied; polling happens via the triggerer,
```
kaxil: Both claims in this comment check out as wrong against the current provider (7.17.0). `deferrable=True` doesn't imply `wait_for_termination=True`: `_handle_deferrable_databricks_operator_execution` gates the `defer()` on `if operator.wait_for_termination:`, so an agent collapsing an existing split and carrying `wait_for_termination=False` over from the old submit task gets fire-and-forget with no polling at all, the exact shape this section is trying to remove.

And `durable=True` checkpoints nothing here: `DatabricksSubmitRunOperator.execute()` branches to `_hook.submit_run()` plus the deferrable helper and never calls `execute_resumable()`, which is the only place the mixin touches `task_state_store`.

Suggest dropping the `durable` half of the comment and making `wait_for_termination=True` explicit in the example. That also softens "Strictly better than the split; no tradeoff to weigh" four lines up, since the deferrable path gives up the run-id checkpoint a `durable=True` submit task has. The Section 2 table row repeats the same "implies" claim and needs the same fix.

skills/airflow-state-store/SKILL.md
```diff
@@ -214,6 +219,7 @@ Use when an operator submits a job to an external system (Spark, Databricks, dbt
 | The task fans out many concurrent I/O operations within a single execution | `async def` task / `BaseAsyncOperator` | Async is for high-throughput I/O, not crash recovery |
 | `retries=0` | — | Crash recovery has nothing to reconnect to |
 | The external system has no trackable job ID (`submit_job` returns `None`) | Plain operator | The mixin's crash-safety guarantee is silently disabled; adds no value |
+| Submit and poll are already split into two separate tasks (submit task → XCom → sensor/poll task) | Nothing — the split already provides it | The "already submitted" checkpoint is the completed upstream task's XCom output, which a retry of the downstream task can't disturb; the downstream task never submits anything, so retrying it is already idempotent. The mixin only earns its keep when submit-then-poll is collapsed into one task's `execute()` call. |
```
kaxil: The reasoning about the downstream sensor task is right, but the conclusion overshoots. The submit task still has the duplicate-submission window: `_hook.submit_run()` returns a run id, worker dies before the task completes, retry submits again.

That is exactly what the mixin covers, and for `DatabricksSubmitRunOperator(wait_for_termination=False)` it's `durable=True` (the default) doing the covering, not the split: `execute()` still routes through `execute_resumable()`, which persists the run id, and `poll_until_complete()` then returns early at the `wait_for_termination` gate. Hand-roll the submit as `@task` plus `DatabricksHook.submit_run()` and there's no protection at all.

Could this say "nothing for the poll task, but the submit task still wants `durable=True`" rather than a flat "Nothing"? As written, the blanket "don't recommend the mixin" on line 212 steers an agent away from the one spot in a split DAG where it does help.

skills/airflow-state-store/SKILL.md
```diff
@@ -302,9 +308,49 @@ class MySparkOperator(BaseOperator, ResumableJobMixin):
...
+                       # not a held worker slot — durable=True (default) still checkpoints the run id
+)
+```
+
+**If no Triggerer is deployed** — this is a genuine tradeoff, not an automatic call either way. It splits into two cases depending on whether holding a worker for the job's duration is acceptable:
+
+- **Job is OK to run synchronously (short/moderate runtime, or worker capacity isn't scarce) and the operator supports durable execution** (inherits `ResumableJobMixin` from Section 5, exposes `durable=True`) — recommend replacing the two-task pattern with **one task** in the operator's durable mode: `wait_for_termination=True`, `durable=True` (both already the default). This is not just fewer moving parts — the operator's built-in poll loop is covered by `ResumableJobMixin` end-to-end (submit *and* poll), so a worker crash mid-poll reconnects instead of resubmitting. A hand-rolled sensor task is a separate, non-mixin code path with its own retry semantics, often weaker (recall the earlier finding that the sensor had no `retries` of its own). Collapsing here removes an entire hand-maintained file, not just a task:
```
kaxil: "recall the earlier finding that the sensor had no `retries` of its own" reads like it leaked in from the review session this was written against. A future reader has no earlier finding to recall, and the failure mode is an agent repeating "the sensor had no retries" to a user as established fact without having looked at the sensor. Maybe "a hand-rolled sensor often has no `retries` of its own"?

skills/airflow-state-store/SKILL.md
```diff
@@ -42,9 +42,12 @@ When the user asks to review a DAG or asks "is there a better way", scan for the
...
 
 Show a before/after snippet when flagging. Use the canonical examples in Steps 3–5 as the "after".
 
+**The submit+sensor split is worth a comment even when the sensor code is bug-free** — reviewing the sensor's code quality (correct `mode=`, correct terminal-state handling, cached hook) is a separate question from whether the two-task split is the right architecture. The full decision tree (Triggerer deployed / no Triggerer + sync-OK / no Triggerer + long-running) lives in one place — **Section 6, "Submit-and-poll DAGs: one task or two?"** — do not re-derive it here; follow that section.
```
kaxil: This says the decision tree lives in one place and not to re-derive it, but the Recommend column four lines up already spells out all three branches. That's two copies to keep in sync, and the table row is the one an agent reaches first. Could the row carry just the "green submit only means accepted" framing and point at Section 6 for the branches?

skills/airflow-state-store/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
+description: Persists task and asset state across retries and DAG runs using Airflow 3.3's AIP-103 key/value stores (`task_state_store`, `asset_state_store`) and the crash-safe `ResumableJobMixin`. Use when the user asks about task state store, checkpointing in tasks, persisting state across retries, job IDs surviving worker crashes, watermarks, asset metadata, resumable tasks, crash-safe operators, or "what's new in Airflow 3.3". Also use proactively when reading a DAG that uses Variables or XCom for intra-task coordination state — flag the anti-pattern and recommend task_state_store or asset_state_store instead. Also use proactively when reviewing ANY DAG that submits a job to an external system and waits for it to finish — Databricks, Snowflake, BigQuery, Redshift, Spark, dbt Cloud, EMR, AWS Batch, etc. — whether that is one submit-and-wait operator or split across a separate submit task plus a sensor/polling task; this covers `wait_for_termination`, `deferrable`, `durable`, hand-rolled sensors polling a run/job id, and whether to collapse a submit+sensor split into one task. Requires Airflow 3.3+.
```
kaxil: "ANY DAG that submits a job to an external system and waits for it to finish" covers close to every production DAG, and it collides with the version gate at the top of the file: below 3.3 the skill is told to say these features aren't available and link the AIP-103 issue instead.

Most of Section 6 isn't 3.3-specific though. `deferrable=True` and `mode="reschedule"` are fine on 3.1, and the Databricks provider even ships an Airflow <3.3 `ResumableJobMixin` stub that accepts `durable=True` and silently always submits fresh. So a 3.2 user reviewing a submit+sensor DAG now trips this skill and gets bounced rather than the advice that does apply. Worth scoping the new trigger to the durable/state-store part, or letting Section 6 answer for pre-3.3.

skills/airflow-state-store/SKILL.md
amoghrajesh: Thanks. I made `wait_for_termination=True` explicit in the example, removed the false durable comment, and softened "no tradeoff to weigh" because the deferred path gives up the run id checkpoint the synchronous `durable=True` path has. I fixed the same "implies" claim in the Section 2 table row too. See my reply on your other comment; that row needed to change anyway.

skills/airflow-state-store/SKILL.md
amoghrajesh: I reworded the row to "Nothing for the poll/sensor task, but the submit task itself still wants durable=True," and updated the scope-check paragraph and Section 6's closing note so neither implies crash-safety stops mattering once the split exists.

skills/airflow-state-store/SKILL.md
amoghrajesh: Fair - that phrasing only made sense inside the review session it was written against. Reworded to "check whether it has its own retries set; many don't," which doesn't ask the reader to recall something that isn't there for them.

skills/airflow-state-store/SKILL.md
amoghrajesh: Agreed. I trimmed the Recommend column to the "green submit ≠ success" framing plus a pointer to Section 6. The three branches, plus the wait_for_termination fix from your other comment, now live in exactly one place.

skills/airflow-state-store/SKILL.md
amoghrajesh: Good catch. I scoped the version gate instead of narrowing the trigger. It now withholds only the `task_state_store`/`asset_state_store`/`durable-crash-safety` claims below 3.3, and states that Section 6's Triggerer-vs-reschedule decision applies on any version. A 3.2 user gets that guidance now instead of getting bounced to the AIP-103 issue link.

amoghrajesh: Handled in 86c14c6

skills/airflow-state-store/SKILL.md
amoghrajesh: Handled in 86c14c6

skills/airflow-state-store/SKILL.md
amoghrajesh: Handled in 86c14c6

skills/airflow-state-store/SKILL.md
amoghrajesh: Handled in 86c14c6

skills/airflow-state-store/SKILL.md
amoghrajesh: Handled in 86c14c6

amoghrajesh:
@kaxil can you take another look when possible please?

kaxil (approved):

PR #261, opened 2026-09-03
Add airflow-custom-provider skill

odaneau-astro:
## Summary
- Adds a skill for packaging a hook/operator/sensor as a pip-installable Airflow provider (`get_provider_info()`, the `apache_airflow_provider` entry point, custom connection types).

## Test plan
- [x] Read through against the canonical template (astronomer/airflow-provider-sample) for drift
- [x] Spot-check the `get_provider_info()` / entry-point verification commands against a real provider install
- [x] Tested the skill end-to-end, generated a provider package, and ran `astro dev init` and `astro dev start`to confirm it installed cleanly.


odaneau-astro:
Manually followed this skill end to end: scaffolded a real provider package, built the wheel, and ran it through \`astro dev init\` / \`astro dev start\` against Astro Runtime \`3.3-7\`.

Found that Step 5's documented local-wheel step doesn't work on current Astro Runtime images — the base image's \`ONBUILD\` chain installs \`requirements.txt\` before it copies the rest of the project (including the wheel) into the image, so a relative path reference fails with \`Distribution not found\`. Latest commit fixes this with an explicit Dockerfile \`RUN\` after \`FROM\`, which I verified resolves cleanly (\`airflow providers list\` / \`airflow providers hooks\` both showed the package and its connection type correctly).

Also added the alternative path for sharing a provider across projects: publish to public/private PyPI and reference it normally in \`requirements.txt\`, including the Airflow-2-vs-3 \`--extra-index-url\`/\`--index-url\` distinction and a pointer to [Astronomer's private Python packages guide](https://www.astronomer.io/docs/cli/v1.44/private-python-packages) for keeping credentials out of git via a build secret.

Separately, this exercise also surfaced that the canonical template this skill points to (astronomer/airflow-provider-sample) is written for Airflow 2 and needs updating for Airflow 3 — filed as astronomer/airflow-provider-sample#20.

PR #263, opened 2026-09-09
fix(diagnostics): report all task failures across MCP and CLI

### LINKED ISSUE #228: diagnose_dag_run misses failed task instances beyond the first 100 (no pagination / no state filter)

AndresElizondo:
Fixes #228.

Failed task instances beyond the first API page were omitted from MCP and CLI diagnostics and trigger-wait results. This change paginates the complete listing before reporting task states or failures, including when Airflow clamps the requested page size.

This builds on #229 and retains Jesus Manuel Gallego's original commit and authorship. It addresses the review feedback there:

- Follow the API's `total_entries`, advance by the actual returned count, and report unavailable/malformed responses, duplicate `(task_id, map_index)` identities, premature empty pages, or the 1,000-page safety limit as errors rather than complete results.
- Apply pagination to both MCP callers and both CLI callers. Preserve the completed run status and expose `failed_tasks_error` when trigger-wait cannot retrieve a complete failure list.
- Share diagnostic summarization and failed-task extraction across MCP and CLI. Include `map_index` consistently so mapped failures remain distinguishable.
- Keep complete state counts and compact failure details while limiting full `task_instances` metadata to the first 100 instances. Expose `task_instances_returned` and `task_instances_truncated`, and document the response behavior.

Regression coverage exercises all four public entry points with 815 instances and a server-imposed 40-instance page size, late mapped failures, interrupted pagination, invalid totals, page limits, and small-run compatibility. An overlapping-page regression reproduces a 101-instance run where an updated index 0 repeats on page two and displaces failed index 100; both adapters and all four callers now report an incomplete listing instead of an empty failure list. Duplicate identities within one page are also covered. This detects overlap but does not provide an atomic snapshot of a changing run. A two-task integration fixture also exercises pagination through both real Airflow APIs.

### Validation

- Python 3.10, 3.11, and 3.12: **660 unit tests passed and 3 pre-existing tests failed on each** with the project's declared dependencies and plugin extra installed. The failures are the three unchanged `TestGetAuthToken` tests; the same failures reproduced on untouched `main`. All pagination regressions pass, including the ten duplicate-identity cases that failed before the guard.
- Airflow 2.11.0 and Airflow 3.1.0: **49 integration tests passed, 1 skipped on each** using the repository's Docker Compose services on an earlier pagination revision; not rerun for the duplicate-identity guard.
- Repository and package `prek run --all-files`: passed, including Ruff, documentation checks, ty, and Bandit. The type checker reports existing warnings but no errors.
- Standalone Docker image build and `python -m astro_airflow_mcp --help` smoke test: passed on an earlier pagination revision.
- No dependencies or lockfile changes.

cc @kaxil for review of this follow-up to #229.


AndresElizondo:
FYI: Fully built by Astra

AndresElizondo:
Tagging a few active contributors: @jlaneve @jeremybeard  :pray:

astro-airflow-mcp/src/astro_airflow_mcp/utils.py
```diff
@@ -77,6 +79,26 @@ def extract_failed_tasks(task_instances: list[dict[str, Any]]) -> list[dict[str,
...
 
 
+def summarize_task_instances(task_instances: list[dict[str, Any]]) -> dict[str, Any]:
+    """Summarize every instance, retaining only a bounded sample of full details.
+
+    Counts and compact failed-task details cover the complete listing. The
+    sample keeps large mapped runs from filling the model's context with full
+    metadata for successful tasks. MCP and CLI diagnostics share this format.
+    """
+    sample = task_instances[:DEFAULT_LIMIT]
```
kaxil, line 89: `[:DEFAULT_LIMIT]` takes the head of the API ordering, so on your own 815-instance fixture in `test_failures_beyond_server_clamped_pages` (813 success, the two failures at map_index 813 and 814) every one of the 100 full-detail instances is a success and neither failure is in there. Sorting failed/upstream_failed to the front of the sample would keep the detail someone diagnosing the run actually wants.

Related: the cap is on the list that doesn't scale with failures, while `summary.failed_tasks` is unbounded. I measured an 815-instance run where one root task fails and 814 go upstream_failed: `failed_tasks` alone serialises to ~178 KB against ~65 KB for the capped sample, so the response ends up bigger than it was before the cap existed. That mass-failure cascade is the case the cap was meant to protect against.

astro-airflow-mcp/src/astro_airflow_mcp/utils.py
```diff
@@ -77,6 +79,26 @@ def extract_failed_tasks(task_instances: list[dict[str, Any]]) -> list[dict[str,
...
+    return {
+        "task_instances": sample,
+        "task_instances_returned": len(sample),
+        "task_instances_truncated": len(sample) < len(task_instances),
+        "summary": {
+            "total_tasks": len(task_instances),
+            "state_counts": dict(Counter(task.get("state", "unknown") for task in task_instances)),
```
kaxil, line 96: The `"unknown"` default only fires when the key is absent, but Airflow sends `"state": null` for task instances that exist without a state yet, which is the normal case on a running DAG. Those count into a `null` key that `json.dumps` writes as `"null"`, so `state_counts` can come back as `{"success": 3, "null": 12}` and, if an instance ever did omit the key, you'd get two buckets for the same condition. `task.get("state") or "unknown"` collapses them.

astro-airflow-mcp/src/astro_airflow_mcp/adapters/base.py
```diff
@@ -430,6 +430,47 @@ def get_task_instances(
...
+            page = self.get_task_instances(dag_id, dag_run_id, limit=page_size, offset=len(merged))
+            batch = page.get("task_instances")
+            if page.get("available") is False or not isinstance(batch, list):
+                raise RuntimeError("Task instance listing unavailable or missing task_instances")
+            total = page.get("total_entries")
+            if type(total) is not int or total < 0:
+                raise ValueError("Invalid total_entries in task instance response")
+            for task in batch:
+                identity = (task["task_id"], task.get("map_index", -1))
+                if identity in seen:
```
kaxil, line 460: Raising here means `diagnose_dag_run` drops `summary` entirely, so a large run that's still moving now returns less than the page-1 counts it used to. Fail loud over silently wrong is the right call, but would a `complete: false` flag on the summary, or one retry of the walk before giving up, get you both? Overlapping pages are normal while a run is actively updating, which is exactly when someone reaches for this tool.

astro-airflow-mcp/tests/integration/test_api_endpoints.py
```diff
@@ -336,6 +336,17 @@ def test_get_task_instances(self, adapter, completed_dag_run):
...
 
+    def test_get_all_task_instances(self, adapter, completed_dag_run):
+        """Pagination collects every instance through the real Airflow 2/3 API."""
+        dag_id, dag_run_id = completed_dag_run
+
+        result = adapter.get_all_task_instances(dag_id, dag_run_id, page_size=1)
+
+        instances = result["task_instances"]
+        assert len(instances) == result["total_entries"]
+        assert len(instances) > 1
```
kaxil, line 347: `completed_dag_run` (conftest.py:69) returns any existing success/failed run it finds before triggering a new one, so against a local stack that stays up between runs this picks a run created before `after_quick_task` was added and `> 1` fails. The `total_entries` and uniqueness assertions around it already prove the `page_size=1` walk worked; skipping stale single-instance runs in the fixture would make this independent of what's already in the metadata DB.

kaxil (approved):
The pagination logic is sound. Counting distinct identities and matching `total_entries` genuinely proves the walk saw every instance, rather than trusting the page count. Docs match the behavior and the test coverage is thorough.

Four non-blocking comments below. The `utils.py` sampling one is the only one I would actually push to get fixed: the cap currently trims the list that does not grow with failures while leaving `failed_tasks` unbounded, so a mass-failure cascade still produces a larger response than before the cap existed.

astro-airflow-mcp/src/astro_airflow_mcp/utils.py
AndresElizondo: On the first part, updated the method to prioritize failed tasks followed by upstream_failed and then other task states.

AndresElizondo: On the second part, I'll follow up on this in another PR to avoid delaying this one further.

astro-airflow-mcp/src/astro_airflow_mcp/utils.py
AndresElizondo: Fixed, defaulting to "unknown"

astro-airflow-mcp/src/astro_airflow_mcp/adapters/base.py
AndresElizondo: Agreed that returning no summary limits usefulness during active runs.
I suggest keeping the explicit error behavior in this PR, we can follow up on a separate PR where we can define retries or partial summaries.

astro-airflow-mcp/tests/integration/test_api_endpoints.py
AndresElizondo: Good catch, fixed by triggering a new run.

AndresElizondo:
Hey @kaxil!
Thanks for the great review. So thorough.
I have pushed a few changes to address some of your simpler suggestions.
We can follow up on the more complicated ones in a new PR.

In the meantime, I'd appreciate another review before we merge. 🙏

PR #271, opened 2026-09-17
Improvement: Add configurable tool allowlist to Airflow MCP

ElianaArjona:
Airflow MCP currently exposes every registered tool, even when an agent only needs a
small subset for its intended workflow. This gives the agent access to capabilities
beyond those the deployment intends it to use.

This change adds `ASTRO_MCP_ALLOWED_TOOLS` so operators can explicitly select the
tools available to an agent. For example, a deployment intended for DAG inspection
can expose:

`ASTRO_MCP_ALLOWED_TOOLS=list_dags,get_dag_details,list_dag_runs,get_dag_run`

Tools outside that list are hidden from discovery, and direct calls to them are
rejected before execution. This enforces the configured capability boundary at the
MCP server, rather than relying on the agent to follow instructions about which tools
it should use.

Leaving the variable unset preserves the existing behavior of exposing all tools.

## Design rationale

Airflow MCP owns tool registration and execution, so it can enforce the same policy
for every connected client.

The allowlist is applied during server startup using FastMCP’s built-in visibility
controls. Tool names are validated before applying the policy: unknown names, empty
values, and empty entries prevent MCP startup. Surrounding whitespace and duplicate
names are ignored.

The setting applies to the entire deployment. Changes require restarting the
processes hosting MCP and refreshing client discovery.

## Validation

- 76 focused tests passed, including discovery filtering, rejection of excluded
calls, and startup configuration validation.
- Lint, formatting, and type checks passed for the changed Python files.
- Verified the standalone HTTP server against a running Airflow instance: discovery
returned exactly the configured tools, an excluded call was rejected, and allowed
tools returned real DAG data.

## How to verify in an Astro development deployment

1. Install a build containing this change.
2. Set `ASTRO_MCP_ALLOWED_TOOLS=list_dags,get_dag_details` on the Airflow processes
hosting the MCP plugin, then restart or redeploy them.
3. Connect an MCP client to `/mcp/v1/` and refresh tool discovery.
4. Confirm that only the selected tools are advertised and can execute with
appropriate Airflow permissions.
5. Send a direct call to `get_variable` and confirm it is rejected as unavailable.
6. Remove the variable, restart the hosting processes, and confirm all tools are
advertised again.

## Scope

The allowlist limits available capabilities; it does not interpret user intent or
restrict arguments passed to an allowed tool. Allowed tools retain their internal
operations.

MCP resources and prompts remain available. Airflow authorization and `AF_READ_ONLY`
continue to apply independently. Per-user or per-channel tool filtering is outside
this change.


ElianaArjona:
Hey @kaxil! Would you be available to review this PR when you have a moment?

It adds ASTRO_MCP_ALLOWED_TOOLS, an optional allowlist so a deployment exposes only the MCP tools an agent actually needs. The reason we need this on our side is security: today, every registered tool is exposed, including ones that can trigger, pause, clear, or delete DAG runs, and we would rather not give an agent that only needs to read DAG state the ability to touch our DAGs. Leaving the variable unset keeps the current behavior.

PR #273, opened 2026-09-21
Fix diagnose_dag_run silently truncating task instances past page 1

markr9805:
## Summary

`diagnose_dag_run` — both the MCP tool (`tools/diagnostic.py`) and the CLI command (`cli/runs.py`) — calls `adapter.get_task_instances(dag_id, dag_run_id)` with no pagination handling. `get_task_instances` defaults to `limit=100, offset=0` against Airflow's `taskInstances` REST endpoint, and neither caller ever inspects the response's `total_entries` or requests a second page.

For any DAG run with more than 100 task instances, everything past the first page is silently dropped — and it doesn't look dropped: `summary.total_tasks` is computed as `len(task_instances)`, so the truncated count is reported as if it were the real total. There's no error, no warning, nothing to indicate anything was cut.

This is easy to hit on a real fleet-style DAG (one task per pod per replication phase, etc.). I ran into it on a production DAG with ~150 task instances per run: `diagnose_dag_run` consistently reported `"total_tasks": 100` and silently omitted roughly a third of the run's tasks, including one I was specifically trying to look up by name (`get_task_instance` for that same task, called directly, returned it fine — it existed and had run successfully).

## Fix

- Added `get_all_task_instances()` in `utils.py`: loops on `total_entries` until every page has been collected, falling back to a single call when `total_entries` is absent (e.g. a minimal adapter response).
- Pointed both `diagnose_dag_run` implementations at it instead of the single unpaginated call.

## Test plan

- [x] New unit tests for `get_all_task_instances` in `tests/test_utils.py`: no-`total_entries`, single-page, multi-page, and empty-run cases.
- [x] New regression test in `tests/test_consolidated_tools.py` driving `diagnose_dag_run` itself through a simulated >100-task run, asserting `total_tasks` and `task_instances` reflect the full run and that pagination offsets are correct.
- [x] Full existing suite passes: `uv run pytest tests/ --ignore=tests/integration` (615 passed).
- [x] `make lint` (ruff check), `make format` (ruff format --check), `make type-check` (ty), and `make security` (bandit) all clean on the changed files.


markr9805:
Maintainer nudge per CONTRIBUTING.md "Request review": could someone approve the workflow run for this fork PR so CI can execute, and take a look when a slot opens? Happy to address anything that comes back.

AndresElizondo:
Hey @markr9805!
I have a PR open that already addresses this issue. Hopefully it'll be merge-worthy soon.
https://github.com/astronomer/agents/pull/263

